PR #3, opened 2025-02-27
Bootstrap CLI

atanasdinov:
Bootstraps a CLI for the new Elemental project.

After careful evaluation, I decided to go with urfave/cli as opposed to Cobra.

Both libraries are wildly popular, however, personally I find the latter really verbose and hard to follow.
It's interesting that in a barebones setup the Cobra CLIs are lighter, however, this quickly changes once you add Viper into the mix. 

Also set up some initial commands that we already know will be necessary:
- build
- install
- version

The first two showcase some flag definitions that I included just for an example. We might need many more flags or rely less on flags and more on configuration files. Regardless, I believe adding in Viper or any other such tooling would make the project convoluted.

Note that the global variables for the flags might seem strange at first glance, however, this is usually what most of the configuration libraries use under the hood as well (e.g. Viper).

Usage examples:
```
» ./elemental -h                    
NAME:
   elemental - Build, install and upgrade infrastructure platforms

USAGE:
   elemental [global options] command [command options]

COMMANDS:
   build       Build new image
   install     Install an OCI image on a target system
   version, v  Inspect program version
   help, h     Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h  show help

» ./elemental install -h 
NAME:
   elemental install - Install an OCI image on a target system

USAGE:
   elemental install [OPTIONS]

OPTIONS:
   --os-image value          OCI image containing the operating system
   --target value, -t value  Target device for the installation process
   --help, -h                show help

» ./elemental install --os-image oci://registry.suse.com/elemental/build-20250227 --target /dev/loop0    
2025/02/27 16:54:42 args: &{OperatingSystemImage:oci://registry.suse.com/elemental/build-20250227 Target:/dev/loop0}
```


davidcassany:
> One of the reason for the brief comments in the action handlers is that it is probably a good idea to separate the actual business logic from the handling of the input i.e. we parse flags, read config files, then run some sort of validation to ensure that these all make sense and only then trigger the main process.

Agree, this was the spirit of Elemental Toolkit too. First collect all inputs and some essential state data (current arch and stuff like that), validate it and after that run the actual big task (install, build, upgrade). This way we assume the task input is already sanitized.

davidcassany:
What I find weird from this approach is having  subcommands definition inside the `pkg`, is there any reason not to include them under `cmd/elemental/<subcommand>`? To my eyes they belong there, but probably I am missing something. Apart from that comment it all looks good to me :+1:

atanasdinov:
This is a good question. My goal was to keep `/cmd` scoped to only the main functions as everything else is most certainly going to grow a lot with time.

Having subpackages under the main package (i.e. the suggested `cmd/elemental/<subcommand>` approach) works to an extent but then the "action" handlers would have to import those subpackages to use the CLI args, eventually error codes, etc. By having as much as possible outside of `/cmd` we are less likely to run into some weird cyclic import issues down the line.

In fact, I believe I should move everything under `/internal` and not `/pkg` because none of the CLI related logic should be reusable outside of this repository. 

WDYT?

davidcassany:
Yes I think `internal` would work for me too.

> Having subpackages under the main package (i.e. the suggested `cmd/elemental/<subcommand>` approach) works to an extent but then the "action" handlers would have to import those subpackages to use the CLI args, eventually error codes, etc. By having as much as possible outside of `/cmd` we are less likely to run into some weird cyclic import issues down the line.

My idea of having everything CLI related in `/cmd` was meant to be the other way round. So having CLI logic having to import form config packages to feed them, this way client logic and all ufavre related imports are under `/cmd` and seen under `/pkg`. So action handlers do not import anything from client args, they just a get full setup which is decoupled form the cli library choice.

atanasdinov:
Alright, now I understand and agree.

It's a little early to tell how this shapes up and I'd like to see how we decide to handle errors (as well as user facing ones).
Let's get this under `/internal` for now and re-evaluate once we have some code in.

I assume we'll frequently move packages around until we stabilise anyway.

davidcassany (approved):
LGTM

PR #4, opened 2025-02-28
Add Golang CI linter job

dirkmueller:

PR #5, opened 2025-03-03
Add Golang/Goheader CI

ldevulder:
Add simple Goheader check. COuld be redondant with the [PR](https://github.com/SUSE/elemental/pull/4) made by @dirkmueller , but as he is on vacation I prefer to open a new one.

Let me know if I should push on the existing PR instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,21 +1,9 @@
-# If you prefer the allow list template instead of the deny list, see community template:
-# https://github.com/github/gitignore/blob/main/community/Golang/Go.AllowList.gitignore
-#
-# Binaries for programs and plugins
-*.exe
-*.exe~
-*.dll
-*.so
-*.dylib
-
-# Test binary, built with `go test -c`
-*.test
-
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-# Dependency directories (remove the comment below to include it)
-# vendor/
+# build and vscode are not needed
+build/
+.vscode/
 
 # Go workspace file
 go.work
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,9 +19,13 @@ linters:
     - dupl # check duplicated code
     - goconst # check strings that can turn into constants
     - gofmt # check fmt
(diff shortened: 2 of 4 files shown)


.golangci.yml
```diff
@@ -0,0 +1,42 @@
...
+    - gofmt # check fmt
+    - goheader # Check license headers, only checks files in current year
+    - goimports # check imports
+    - gocyclo # check complexity
+linters-settings:
+  gocyclo:
+    min-complexity: 20 # default is 30 which is too high
+  goheader:
+    template: |-
+      Copyright © 2022 - {{year}} SUSE LLC
```
ldevulder, line 30: As some code is already taken from Elemental with 2022 as the base year I think it's fine to keep it as is.

.gitignore
```diff
@@ -1,21 +1,6 @@
...
-# Test binary, built with `go test -c`
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
-# Dependency directories (remove the comment below to include it)
-# vendor/
+# build and vscode are not needed
+build/
```
atanasdinov, line 2: Can we keep the previous diff too?

.golangci.yml
```diff
@@ -0,0 +1,42 @@
+run:
+  timeout: 5m
+  tests: false
+output:
+  formats:
+    - format: github-actions
+    - format: colored-line-number
+linters:
+  disable-all: true
+  enable:
```
atanasdinov, line 10: Can we add the following to the list:
- gocritic
- gosec
- unparam
- errorlint

.gitignore
ldevulder: You mean keeping the `*.exe|*.test|*.out` stuff as well?

atanasdinov: Keeping `*.out` for sure. The rest we can live with -- the .gitignore file currently follows a generic template.

ldevulder: Ok, I will read it. `exe|test` were not very useful, that's why I removed them ;-)

atanasdinov (approved):

PR #6, opened 2025-03-03
Add SUSE/elemental-team as codeowners

frelon:


ldevulder:
Was done in the PR I opened this morning: https://github.com/SUSE/elemental/pull/5 :smiley:

atanasdinov (approved):
@ldevulder I'd suggest to merge this and then your PR can be rebased. The latter has a single commit with both this as well as many other things so I think it's better to keep the history clean.

ldevulder (approved):

PR #7, opened 2025-03-05
Add missing Goheaders

ldevulder:
Add missing headers in Go files to make `golangci-lint` happy.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,7 +31,7 @@ linters-settings:
     min-complexity: 20 # default is 30 which is too high
   goheader:
     template: |-
-      Copyright © 2022 - {{year}} SUSE LLC
+      Copyright © {{year-range}} SUSE LLC
 
       Licensed under the Apache License, Version 2.0 (the "License");
       you may not use this file except in compliance with the License.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,6 +32,7 @@ linters-settings:
   goheader:
     template: |-
       Copyright © {{year-range}} SUSE LLC
+      SPDX-License-Identifier: Apache-2.0
 
       Licensed under the Apache License, Version 2.0 (the "License");
       you may not use this file except in compliance with the License.
--- a/cmd/elemental/main.go
+++ b/cmd/elemental/main.go
@@ -1,3 +1,20 @@
+/*
+Copyright © 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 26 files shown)


atanasdinov (approved):
Can we use 2025 only everywhere? 👀

ldevulder:
> Can we use 2025 only everywhere?

The issue is that some code will be copy/paste from current Elemental code (see PR #2) and not sure if the linter will be happy. Will check.

atanasdinov:
You're right. I guess maybe we can actively adjust the code to point to 2025 instead? The linter will fail unless the new files are adjusted and I guess it makes sense to go for 2025 even if some of the code is completely copy-pasted. Thoughts?

ldevulder:
Using `{{year-range}}` should be fine for almost all cases, I will fix all.

fdegir:
I have two comments about this, not specific to this PR or the files in this repository but a general approach to copyright notices, license files and so on. Please note that I'm not sure if SUSE has a policy around this and if so, what I suggest below may not be valid.

* There is a movement to drop years in the license headers in individual files to reduce the maintenance burden and avoid inconsistencies. [1][2][3]
* I suggest using SPDX compliant license IDs in the headers to allow license information to be easily extracted by scripts or code scanning tools. [4]

[1] https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects
[2] https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/
[3] https://hynek.me/til/copyright-years/
[5] https://spdx.dev/learn/handling-license-info/

davidcassany:
> @davidcassany the internal `{{year-range}}` check uses the syntax `year1-year2` and not the one we have in older Elemental code which is `year1 - year2`... So I can create a regex to match the old behavior or we can fix the copied code in the PR with a `sed` for example, WDYT is the best/easiest?

I'd simply fix copied code and adapt wherever necessary. In any case if we can avoid having the year there that would be even better. So I think we should try to figure out the actual requirements for the header first.

ldevulder:
> * There is a movement to drop years in the license headers in individual files to reduce the maintenance burden and avoid inconsistencies. [1][2][3]

Yes but legal team asked for this some times ago. See ou Slack channel for a more detailed answer.

> * I suggest using SPDX compliant license IDs in the headers to allow license information to be easily extracted by scripts or code scanning tools. [4]

I was thinking of this yes, but we followed the Apache license [rules](https://www.apache.org/licenses/LICENSE-2.0#apply), but I think we can add it without any issue. Any comment from others on this? I can add it anyway.

ldevulder:
> So I think we should try to figure out the actual requirements for the header first.

More detailed answer in the Slack channel, but it seems that we should keep it as-is.

davidcassany (approved):
LGTM

PR #9, opened 2025-03-05
Add cleanstack and chroot utilities

davidcassany:
This commits adds the chroot and cleanstack utilities from former Elemental Toolkit.


pkg/utils/cleanstack/clean_stack.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	errorOnly = iota
+	successOnly
+	always
+)
+
+type CleanFunc func() error
+
+// CleanJob represents a clean task. In can be of three different types. ErrorOnly type
```
fgiudici, line 30:
```suggestion
// CleanJob represents a clean task. It can be of three different types: errorOnly type
```

fgiudici (approved):
LGTM

pkg/utils/chroot/chroot.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// Prepare will mount the defaultMounts as bind mounts, to be ready when we run chroot
+func (c *Chroot) Prepare() error {
+	var err error
+	keys := []string{}
+	mountOptions := []string{"bind"}
+
+	if len(c.activeMounts) > 0 {
+		return errors.New("There are already active mountpoints for this instance")
```
atanasdinov, line 83:
```suggestion
		return errors.New("there are already active mountpoints for this instance")
```

pkg/utils/cleanstack/clean_stack.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	errorOnly = iota
+	successOnly
+	always
+)
+
+type CleanFunc func() error
+
+// CleanJob represents a clean task. It can be of three different types: errorOnly type
```
atanasdinov, line 30: What does "clean task" mean i.e. is the word "clean" necessary everywhere?

atanasdinov (commented):
Can we drop the "utils" package that these fall under? This is a little tricky but "utils" is always so generic that it basically serves no specific purpose.

atanasdinov (approved):

PR #11, opened 2025-03-05
Add Unit Test execution workflow

davidcassany:
This commit adds a Unit Tests workflow to run unit tests and it also checks code coverage does not go below the 70% threshold.


ldevulder:
Do you think that a badge could be useful?

atanasdinov (approved):

PR #15, opened 2025-03-06
Adding additional FS utilities

davidcassany:
This commit adds additional utilities to resolve symlinks, including nested symlinks.

Moreover this commit adds utilities to find files based on given root and pattern.


atanasdinov (approved):

fgiudici (approved):
Looks good here

PR #16, opened 2025-03-06
Adding diskrepart package

davidcassany:
This diskrepart package defines objects and interfaces to handle disk partitioning tasks. There are two different implementations for the partitioner interface, one based on parted and another one based on sgdisk.

Former Elemental Toolkit started wrapping around parted to provide legacy msdos partition table support. However to build RAW disks in unprivileged environments sgdisk was required and a new implementation of the interface got included.


fgiudici (approved):
LGTM

PR #17, opened 2025-03-06
Extend runner interface and add rsync package

davidcassany:
This commit extends Runner interface with a couple of additional methods to support running commands with an associated Context. It also provides a new method to parse in real-time stdout and stderr.

The rsync package is wrapper for rsync command. It can be initiated with a Context in case the caller considers some sort of cancellation logic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-GINKGO?= "github.com/onsi/ginkgo/v2/ginkgo"
+GINKGO?="github.com/onsi/ginkgo/v2/ginkgo"
 
-PKG:=./pkg/...
+PKG?=./pkg/...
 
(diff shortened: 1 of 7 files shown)


davidcassany:
@frelon this is adding a couple of new functions in runner interface to lunch commands within a Context. My understanding of it is this is worth for commands that might take a while, such as rsync, setfiles, etc. So we can cancel them. I doubt it is worth for immediate commands.

With this can easily incorporate context handling at a higher level and consider handling signals within the client.

frelon (approved):
Nice, I like the new runner interface! :+1:

fgiudici (approved):
LGTM

PR #18, opened 2025-03-06
Adds unpacker package

davidcassany:
The unpacker package provides an interface to unpack object including a root-tree to a specific location.

This implements the OCI unpacker, but we could easily consider additional transports such as tarballs, filesystem images, etc.


fgiudici (approved):
Looks good!

frelon (approved):

PR #20, opened 2025-03-18
Add dependabot config

frelon:
Add config for dependabot to update direct go.mod dependencies and
github-actions weekly.

The PRs will be opened with the "dependencies" label.


dirkmueller (approved):
Wfm. might be worth considering scheduling more frequently for go module updates but that can still be done later.

atanasdinov (approved):

PR #28, opened 2025-03-20
Add deployment package

davidcassany:
This commit adds the deployment package. This package aims to structure all the information required by a deployment and its sanity checks.

In addition this commit also includes initial logic to feed the install command with the deployment setup.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,10 +13,10 @@ require (
 	github.com/twpayne/go-vfs/v4 v4.3.0
 	github.com/urfave/cli/v2 v2.27.6
 	k8s.io/mount-utils v0.32.3
+	sigs.k8s.io/yaml v1.4.0
 )
 
 require (
-	github.com/Azure/go-ansiterm v0.0.0-20250102033503-faa5f7b0171c // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/Microsoft/hcsshim v0.12.9 // indirect
 	github.com/StackExchange/wmi v1.2.1 // indirect
--- a/internal/cli/action/install_test.go
+++ b/internal/cli/action/install_test.go
@@ -1,0 +1,110 @@
+/*
+Copyright © 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package action_test
+
+import (
+	"bytes"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/suse/elemental/v3/internal/cli/action"
+	"github.com/suse/elemental/v3/internal/cli/cmd"
+	"github.com/suse/elemental/v3/pkg/log"
+	"github.com/suse/elemental/v3/pkg/sys"
+	sysmock "github.com/suse/elemental/v3/pkg/sys/mock"
+)
+
+const (
+	badConfig = `
+disks:
+- partitions:
+  # no system partition
+  - label: mylabel
+    function: data
+    fileSystem: xfs
+  - function: efi
+    fileSystem: vfat
+  - function: recovery
+    fileSystem: btrfs
+  - function: data
+    fileSystem: ext2
+`
+)
+
+var _ = Describe("Install action", Label("install"), func() {
+	var s *sys.System
+	var tfs sys.FS
+	var cleanup func()
+	var err error
+	var ctx *cli.Context
+	var buffer *bytes.Buffer
+
+	BeforeEach(func() {
+		cmd.InstallArgs = cmd.InstallFlags{}
+		buffer = &bytes.Buffer{}
+		tfs, cleanup, err = sysmock.TestFS(map[string]string{
+			"/configDir/bad_config.yaml": badConfig,
+			"/dev/device":                "device",
+		})
+		Expect(err).NotTo(HaveOccurred())
+		s, err = sys.NewSystem(
+			sys.WithFS(tfs),
+			sys.WithLogger(log.New(log.WithBuffer(buffer))),
+		)
+		Expect(err).NotTo(HaveOccurred())
+		ctx = cli.NewContext(cli.NewApp(), nil, &cli.Context{})
+		if ctx.App.Metadata == nil {
+			ctx.App.Metadata = map[string]any{}
+		}
+		ctx.App.Metadata["system"] = s
+	})
+
+	AfterEach(func() {
+		cleanup()
+	})
+	It("fails if no sys.System instance is in metadata", func() {
+		ctx.App.Metadata["system"] = nil
+		Expect(action.Install(ctx)).NotTo(Succeed())
+	})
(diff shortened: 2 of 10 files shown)


davidcassany:
This deployment package is expected to grow and also include extensions, charts and any arbitrary data it might require (tarballs?). It also is missing the bootloader and snapshotter setup at this point.

My idea is that the deployment structure is the entry point of `build`, `install`, `upgrade` and eventually `reset` logic. Which doesn't necessarily mean this is the UI for all of these (e.g. disks layout details are required to upgrade but not configurable)

pkg/deployment/deployment.go
```diff
@@ -0,0 +1,379 @@
...
+
+type Disk struct {
+	Device     string       `json:"device,omitempty"`
+	Partitions []*Partition `json:"partitions"`
+	SectorSize uint         `json:"sectorSize,omitempty"`
+}
+
+type Deployment struct {
+	SourceOS string  `json:"sourceOS"`
+	Disks    []*Disk `json:"disks"`
```
davidcassany, line 180: Considering multiple disks, I think it shouldn't make the install logic more complicated and it would allow setups like having `/var` or any other volume in a separate disk (probably for performance reasons, crypto...). Adding such a feature in the future without this structure here would be complex I think.

davidcassany (commented):
For simplicity and to start fast I have not considered the definition of a schema, I wonder if it would make sense. I'd like having a schema of the deployment, but I also don't want to over complicate things. Also json schemas do not support co-ocurrence validation AFAIK.

go.mod
```diff
@@ -10,14 +10,13 @@ require (
 	github.com/onsi/ginkgo/v2 v2.23.0
 	github.com/onsi/gomega v1.36.2
 	github.com/sirupsen/logrus v1.9.3
-	github.com/suse/elemental v0.0.0-20250228142830-404c12c956c8
```
frelon, line 13: Rebase on top of https://github.com/SUSE/elemental/pull/21 instead.

frelon (changes requested):

davidcassany:
Rebased on top of #21

pkg/deployment/deployment.go
```diff
@@ -0,0 +1,379 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/suse/elemental/v3/pkg/sys"
+)
+
+const (
+	EfiLabel = "EFI"
+	EfiMnt   = "/boot/efi"
```
frelon, line 29: ESP should be mounted at `/boot` (or `/efi`) according to [BLS](https://uapi-group.org/specifications/specs/boot_loader_specification/#mount-points)

pkg/deployment/deployment.go
```diff
@@ -0,0 +1,379 @@
...
+	EfiSize  = 1024
```
frelon, line 30: We should probably specify a unit in the name, or use a type that converys the units used.

pkg/deployment/deployment.go
```diff
@@ -0,0 +1,379 @@
...
+	RecoverySize  = 0
+
+	SystemLabel      = "SYSTEM"
+	SystemMnt        = "/"
+	AllAvailableSize = -1
+
+	Unknown = "unknown"
+)
+
+type PartFunction int
```
fgiudici, line 43: As a SWE, when I see "Function" I get biased on the meaning 😅 ... what if we use "PartKind" or "PartRole" ?

internal/cli/action/install_test.go
```diff
@@ -0,0 +1,109 @@
...
+	})
+	It("fails to start installing if the configuration file can't be read", func() {
+		cmd.InstallArgs.Target = "/dev/device"
+		cmd.InstallArgs.OperatingSystemImage = "my.registry.org/my/image:test"
+		cmd.InstallArgs.ConfigFile = "doesntexist"
+		err = action.Install(ctx)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("config file 'doesntexist' not found"))
+	})
+	It("fails to start installing if the target device does not exist", func() {
```
atanasdinov, line 94: This test fails with `<string>: config file 'doesntexist' not found: Lstat doesntexist: operation not permitted` locally. Our Makefile (`unit-tests` target) seems to be incomplete.

internal/cli/cmd/root.go
```diff
@@ -21,19 +21,53 @@ import (
...
+			return err
+		}
+
+		if ctx.Bool("debug") {
+			s.Logger().SetLevel(log.DebugLevel())
+		}
+		if ctx.App.Metadata == nil {
+			ctx.App.Metadata = map[string]any{}
+		}
+		ctx.App.Metadata["system"] = s
```
atanasdinov, line 68:
```suggestion
		ctx.Context = context.WithValue(ctx.Context, "system", s)
```

I'm wondering if the context itself makes a better use of passing this value around. It is more common and I can see both working so I'm just leaving the idea if you like it.

pkg/deployment/deployment.go
```diff
@@ -0,0 +1,379 @@
...
+	RWVolumes   []RWVolume   `json:"rwVolumes,omitempty"`
+}
+
+type Disk struct {
+	Device     string       `json:"device,omitempty"`
+	Partitions []*Partition `json:"partitions"`
+	SectorSize uint         `json:"sectorSize,omitempty"`
+}
+
+type Deployment struct {
```
atanasdinov, line 178: I don't see such term in the existing `elemental-toolkit` codebase. Could you please explain why it is picked here?

P.S. Note that I'm not against calling it a "Deployment", I'm only looking to understand the idea behind it.

atanasdinov (commented):
As mentioned offline, great work!

I left a couple of comments and I'm ready to approve once we discuss them.

My only concern is around the validation itself. I understand the idea behind setting sensible values if the user is incorrect but it might make sense to reject improper input based on a validation approach. This is not something to address in this PR obviously but I felt like saying it out loud.

pkg/deployment/deployment.go
davidcassany: I am not really good at naming. I used `Deployment` because it describes the OS deployment on that node and I did not want to call it `Install` as I expect to use it for `upgrade`, `build` and  `reset` (if ever consider it). My idea is that after the installation this is serialized and stored in `/etc/elemental/deployment.yaml` (or a similar file) so this can be easily loaded in `upgrade` process to get the current system state, preventing the need of running heuristics to figure out partition labels, UUIDs, RW paths, etc.

internal/cli/cmd/root.go
davidcassany: Since this is just used to workaround urfave I'd prefer to stick with the urfave utilites for that matter (this is just passing data from root command to subcommands). The Context inside the urfave context is likely to go down deeper in the code and there is not actually needed a context with values. In former `pkg/sys` PR it was already discussed as a preferred approach passing data as explicit arguments.
So to ensure the System is not pulled from the context down the road I think it is just better to not include it there. The `App.Metadata` will only be used in the `internal/cmd` only.

internal/cli/action/install_test.go
davidcassany: I see this might be a race condition, ginkgo randomizes and parallelizes tests order, I guess clearing command args in each test is required. The test passes on my local env and in GH action, but probably fails sometimes depending on the ginkgo concurrence. I am having a look.

davidcassany: Not a race condition, in fact, using a parallel tests I could not reproduce it on my env as each tests runs in its own process.  The problem raised using sequential tests, the cmd.InstallArgs global variable needs to be cleared in each test in that case.

frelon (approved):

fgiudici (approved):

atanasdinov (approved):

davidcassany:
Rebased on a single commit before merging

pkg/deployment/deployment.go
ldevulder: Maybe `OSDeployment` in that case?

PR #33, opened 2025-03-25
Implement version command

frelon:
This commit implements the version command in the elemental binary.

The output looks like the following with a git tag and then commit
suffix:

```
make elemental
./elemental version
v0.1.2+g19c84c9
```


fgiudici (approved):
LGTM

Makefile
```diff
@@ -2,6 +2,20 @@ GINKGO?="github.com/onsi/ginkgo/v2/ginkgo"
 
 PKG?=./pkg/...
 
+GO_MODULE ?= $(shell go list -m)
+
+GIT_COMMIT?=$(shell git rev-parse HEAD)
+GIT_COMMIT_SHORT?=$(shell git rev-parse --short HEAD)
+GIT_TAG?=$(shell git describe --candidates=50 --abbrev=0 --tags 2>/dev/null || echo "v0.0.1" )
+VERSION?=$(GIT_TAG)-g$(GIT_COMMIT_SHORT)
```
atanasdinov, line 10: Do we want to use this somewhere?

internal/cli/cmd/version.go
```diff
@@ -21,14 +21,26 @@ import (
 	"fmt"
 
 	"github.com/urfave/cli/v2"
+
+	"github.com/suse/elemental/v3/internal/version"
```
atanasdinov, line 25: The initial idea in mind was for the `cmd` package to never import anything but `urfave/cli` types and functions. This way it seemed cleaner to me. Do we benefit from inlining the action in this module?

internal/version/version.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var (
+	version = "v0.0.1"
+	// gitCommit is the git sha1
+	gitCommit = ""
+)
+
+// BuildInfo describes the compile time information.
+type BuildInfo struct {
```
atanasdinov, line 31: Out of curiosity, what do you think about reading this data from the `runtime/debug` package directly?

frelon: I did not find a way to read the tag from `runtime/debug`

internal/cli/cmd/version.go
frelon: Resolved in latest version (removed `internal/version` package)

Makefile
frelon: Good catch, removed :+1:

davidcassany (approved):
LGTM

ipetrov117 (approved):

internal/version/version.go
atanasdinov: This is usually the case because you can have multiple tags on the same commit. I'm good with the simplified version, it's nice!

atanasdinov (approved):

PR #34, opened 2025-03-27
Refactor few details of the utility packages

davidcassany:
This commit refactors some of the utility packages:

* Extends mounter interface to allow listing all mountpoints for a given device. This way we can get all mounted btrfs subvolumes for a given btrfs partition.
* Extends the block device package to also provide the filesystem UUID information for the detected partitions. This way we can easily handle partitions and fstab by UUID.
* Extends disk partitioning package to also include UUID as a parameter for formatting. So we can define UUIDs for filesystem


pkg/sys/mounter/mounter.go
```diff
@@ -17,18 +17,26 @@ limitations under the License.
...
+	// paths. The returned slice does not include the given path.
+	GetMountRefs(pathname string) ([]string, error)
+	// GetMountPoints parses /proc/mounts or equivalent data to fetch all available mountpoints for the given device
+	GetMountPoints(device string) ([]MountPoint, error)
 }
 
-func NewDummyMounter() mount.Interface {
-	return mount.NewFakeMounter([]mount.MountPoint{})
+// MountPoint represents a single line in /proc/mounts or /etc/fstab.
+type MountPoint struct {
```
davidcassany, line 37: Since the interface interacts with this new type the interface definition had to be moved to a separate package to prevent circular dependencies. That's why there is  a nested package in mounter just for the implementation.

pkg/sys/mounter/mounter.go
```diff
@@ -17,18 +17,26 @@ limitations under the License.
...
+type Interface interface {
+	Mount(source string, target string, fstype string, options []string) error
+	Unmount(target string) error
+	// IsMountPoint check /proc/mounts or equivalent data to check if the given path is listed there
+	IsMountPoint(path string) (bool, error)
+	// GetMountRefs finds all mount references to pathname, returning a slice of
+	// paths. The returned slice does not include the given path.
+	GetMountRefs(pathname string) ([]string, error)
+	// GetMountPoints parses /proc/mounts or equivalent data to fetch all available mountpoints for the given device
+	GetMountPoints(device string) ([]MountPoint, error)
```
davidcassany, line 33: This method is not part of the original k8smount interface, hence the mounter implementation required a wrapper type.

pkg/block/lsblk/lsblk.go
```diff
@@ -52,14 +53,15 @@ type jParts []*block.Partition
...
-		FS:              p.FS,
-		Flags:           []string{},
-		MountPoints:     p.MountPoints,
-		Path:            p.Path,
-		Disk:            p.Disk,
-		Name:            p.Name,
+		Label:       p.Label,
+		Size:        uint(p.Size / (1024 * 1024)),
+		FileSystem:  p.FS,
+		UUID:        p.UUID,
```
davidcassany, line 59: This is the filesystem UUID which could be used in fstab as `UUID=afadfa` to indentify the device.

pkg/block/device.go
```diff
@@ -35,14 +35,15 @@ type Device interface {
...
-	Name            string
-	FilesystemLabel string
-	Size            uint
-	FS              string
-	Flags           []string
-	MountPoints     []string
-	Path            string
-	Disk            string
+	Name        string
+	Label       string
```
davidcassany, line 39: Renamed FilesystemLabel to Label as this is the common notation in linux in fstab or kernel args (e.g. `LABEL=SYSTEM`) to identify the target device.

pkg/diskrepart/mkfs.go
```diff
@@ -60,6 +85,10 @@ func (mkfs MkfsCall) buildOptions() ([]string, error) {
 			opts = append(opts, "-n")
 			opts = append(opts, mkfs.label)
 		}
+		if mkfs.uuid != "" {
+			opts = append(opts, "-i")
+			opts = append(opts, strings.Split(mkfs.uuid, "-")[0])
```
davidcassany, line 90: VFat filesyste does not really have UUIDs, however it is a common practice to use the VolumeID as such. The VFat volume ID is represented by 8 hex values, like the first chunk of UUID bytes, so for ease of use, to define the vfat volume ID a valid UUID is required and only the first 8 bytes will be used to defined the volumeID.

fgiudici (approved):

pkg/sys/mounter/mounter.go
davidcassany: This is no longer the case

frelon (approved):

PR #35, opened 2025-03-27
Adds ImageSource type

davidcassany:
This commit adds image source type which is used to handle image sources as uri where the uri scheme defines the source type.

It currently supports oci, dir and raw. For instance:

* "raw:///path/to/image.raw"
* "dir:///path/to/directory/tree"
* "oci://my.registry.org/image:tag"


fgiudici (approved):
Looks good here 👍🏼

PR #36, opened 2025-03-28
Keep the FS interface and vfs together

davidcassany:
This commit moves all FS logic and interface definition to vfs package. This keeps package purpose easier to read and prevents having a sys package with vague purpose.

Sys package is just a collection of certain system utilities and the fact they are interfaces is just an encapsulation to allow easy testing and to enforce a proper encapsulation in code.

We are not aiming to build a sys library that can be imported without forcing to also import specific implementations of the interfaces it uses.


davidcassany:
Rebased

fgiudici (approved):
Looks good here

frelon (approved):

PR #39, opened 2025-04-02
Handle deployment file

davidcassany:
This commit includes utilities to write and read /etc/elemental/deployment.yaml file, which is essentially a serialization of the Deployment struct.

This is helpful for life cycle management to know the details of a current system.

In addition, paritions size type has been changed from int to uint.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -24,24 +24,28 @@ import (
 	"sort"
 	"strings"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/suse/elemental/v3/pkg/sys"
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
 )
 
-type MiB int64
+type MiB uint
 
 const (
 	EfiLabel     = "EFI"
 	EfiMnt       = "/boot"
 	EfiSize  MiB = 1024
 
-	RecoveryLabel = "Recovery"
+	RecoveryLabel = "RECOVERY"
 	RecoveryMnt   = "/run/elemental/recovery"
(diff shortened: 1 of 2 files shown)


fgiudici (approved):
LGTM 👍🏼 
Question from a functional PoV: we will have the /etc/elemental/deployment.yaml which looks like a configuration file that can be edited: do we allow editing it? Will it be supported, meaning at OS update time the disk layout could be changed?

davidcassany:
> LGTM 👍🏼 Question from a functional PoV: we will have the /etc/elemental/deployment.yaml which looks like a configuration file that can be edited: do we allow editing it? Will it be supported, meaning at OS update time the disk layout could be changed?

Good point, my idea is that it **cannot** be edited and it is just a place to store the install/upgrade time setup. Probably we could live without to be honest, but then we probably need a way to detect the setup (parse fstab to get UUIDs and RW paths, logic to determine the role of the partition, etc.), just to feed again a deployment struct. See it as the state.yaml file we have in Elemental Toolkit. To make it obvious this is not expected to be touched probably we can consider some different format around it like base64 coding?

davidcassany:
> A banner with "# self-generated, do not edit" and chmod 444 it 🤷🏼 ?

This is a valid an easy first step, I am on it :+1:

dirkmueller (approved):

ldevulder:
> A banner with "# self-generated, do not edit" and chmod 444 it 🤷🏼 ?

Having some kind of checksum as well in the comment to validate the file could be added as well? Without the checksum comment itself of course, to avoid chicken/egg problem :D

PR #41, opened 2025-04-03
Adding file bind mounts to chroot

davidcassany:
This commit extends chroot package to also support bind mounts of files. If the file does not exist in the new root it creates a dummy file to bind mount to it. Any created file to bind mount is removed on chroot close.


dirkmueller (approved):
LGTM. I'd appreciate though if the Error() casing warnings would be handled in a separate commit though.

fgiudici (approved):
LGTM

PR #45, opened 2025-04-03
Few improvements

davidcassany:
This PR includes few fixes and improvements I spot on while running local tests.

Since most of those are one liner changes I collected them in a single PR in separate commits

### DIFF
--- a/pkg/block/device.go
+++ b/pkg/block/device.go
@@ -86,7 +86,7 @@ func (pl PartitionList) GetByUUID(uuid string) *Partition {
 
 	for _, p := range pl {
 		if p.UUID == uuid {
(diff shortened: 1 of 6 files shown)


dirkmueller (approved):

pkg/block/device.go
```diff
@@ -86,7 +86,7 @@ func (pl PartitionList) GetByUUID(uuid string) *Partition {
 
 	for _, p := range pl {
 		if p.UUID == uuid {
-			return part
+			return p
```
fgiudici, line 89: good catch

fgiudici (approved):
LGTM

PR #46, opened 2025-04-04
Implement unpack-image command

frelon:
Unpack-image unpacks a remote or local image into a target directory.

### DIFF
--- a/internal/cli/action/unpack.go
+++ b/internal/cli/action/unpack.go
@@ -18,8 +18,9 @@ limitations under the License.
 package action
 
 import (
-	"context"
 	"fmt"
+	"os/signal"
+	"syscall"
 
 	"github.com/urfave/cli/v2"
 
@@ -43,12 +44,21 @@ func Unpack(ctx *cli.Context) error {
 		WithPlatformRef(args.Platform).
 		WithVerify(args.Verify)
 
-	_, err := unpacker.Unpack(context.Background(), args.TargetDir)
+	ctxSignal, stop := signal.NotifyContext(ctx.Context, syscall.SIGTERM, syscall.SIGINT)
+	defer stop()
+
+	go func() {
+		<-ctx.Done()
+		stop()
+	}()
+
+	_, err := unpacker.Unpack(ctxSignal, args.TargetDir)
 	if err != nil {
 		s.Logger().Error("Failed to unpack image %s: %s", args.Image, err.Error())
+		return err
 	}
 
 	s.Logger().Info("Image %s unpacked", args.Image)
 
-	return err
+	return nil
 }
--- a/internal/cli/action/unpack.go
+++ b/internal/cli/action/unpack.go
@@ -1,0 +1,54 @@
+/*
+Copyright © 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package action
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


davidcassany:
@frelon I guess this conflicting with #42, it should be easy to rebase though.

Nice porting this command, it is a handy one 😉

frelon:
> @frelon I guess this conflicting with #42, it should be easy to rebase though.
> 
> Nice porting this command, it is a handy one 😉

Yep, I put it up as a draft quickly to share, it's also missing tests but I will rebase and add that once #42 is merged :+1:

internal/cli/cmd/unpack-image.go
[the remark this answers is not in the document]
frelon: In elemental we used to name the `cmd/*` files exactly as the actual command so it was clear where the command was implemented.

I'm not against changing to use _ but maybe we should have that in a style-guide and a linter error so it's not subjective.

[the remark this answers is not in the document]
rdoxenham: Ahh got it, my mistake/misunderstanding. Thanks for clearing this up! A style-guide/linter would certainly be a good outcome.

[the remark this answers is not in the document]
frelon: I will bring this up on our weekly to see if we add this as a lint-rule.

internal/cli/action/unpack.go
```diff
@@ -0,0 +1,54 @@
...
+	var s *sys.System
+	args := &cmd.UnpackArgs
+	if ctx.App.Metadata == nil || ctx.App.Metadata["system"] == nil {
+		return fmt.Errorf("error setting up initial configuration")
+	}
+	s = ctx.App.Metadata["system"].(*sys.System)
+
+	s.Logger().Info("Starting unpack action with args: %+v", args)
+
+	unpacker := unpack.NewOCIUnpacker(s, args.Image).
```
davidcassany, line 41: I am wondering if it would make sense here to use the unpacker interface to support the same sort of types as in install command. The only detail to consider are the specific OCI options which are not existing within the other unpacker types.

In fact I realize that Local, Platform and Verify options are not available at install time as they are not mapped to anything. Probably it makes sense trying to fix this first and then rebase again and use the `NewUnpacker` method instead. I guess we can make these options generic and only effective in OCI case.

internal/cli/action/unpack.go
```diff
@@ -0,0 +1,54 @@
...
+		WithLocal(args.Local).
+		WithPlatformRef(args.Platform).
+		WithVerify(args.Verify)
+
+	_, err := unpacker.Unpack(context.Background(), args.TargetDir)
```
davidcassany, line 46: I'd use the context of the client here and probably set a signal handling to trap interrupts as in install command. The whole idea here was to be able to gracefully stop commands and since unpacker interface makes use of a context I think we should use it here.

davidcassany (changes requested):
I'd essentially add a signal management to trap interrupts as done in install command.

Regarding the support of other types of image sources I think we can evolve in a second iteration.

internal/cli/action/unpack.go
frelon: Nice catch, I added the signal handling now.

davidcassany (approved):
LGTM

internal/cli/action/unpack.go
```diff
@@ -0,0 +1,63 @@
...
+	defer stop()
+
+	go func() {
+		<-ctx.Done()
+		stop()
+	}()
+
+	_, err := unpacker.Unpack(ctxSignal, args.TargetDir)
+	if err != nil {
+		s.Logger().Error("Failed to unpack image %s: %s", args.Image, err.Error())
```
atanasdinov, line 57: We need to return after this, otherwise we'd log the successful message below.

atanasdinov (approved):

internal/cli/action/unpack.go
frelon: Nice catch!

pkg/unpack/oci.go
```diff
@@ -48,34 +48,33 @@ type OCI struct {
 
 type OCIOpt func(*OCI)
```
ldevulder, line 49: In fact I just saw that this `type` is never used (was before the changes made in this PR). Maybe we could remove it?

frelon: Nice catch, I will put up a new PR!

PR #47, opened 2025-04-04
Add selinux package

davidcassany:
This PR adds the SELinux package (a setfiles wrapper to relabel a directory tree) and extends the chroot package to allow setting chroot envs without the default bind mounts (/dev, /proc, /sys), which are not always required.

### DIFF
--- a/pkg/selinux/selinux.go
+++ b/pkg/selinux/selinux.go
@@ -31,6 +31,7 @@ const (
 	SelinuxTargetedContextFile = selinuxTargetedPath + "/contexts/files/file_contexts"
 
 	selinuxTargetedPath = "/etc/selinux/targeted"
+	debugLines          = 10
 )
 
 // Relabel will relabel the system if it finds the context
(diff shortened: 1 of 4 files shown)


pkg/selinux/selinux.go
```diff
@@ -0,0 +1,119 @@
...
+func Relabel(ctx context.Context, s *sys.System, rootDir string, extraPaths ...string) error {
+	contextFile := filepath.Join(rootDir, SelinuxTargetedContextFile)
+	contextExists, _ := vfs.Exists(s.FS(), contextFile)
+
+	if contextExists {
+		var err error
+		args := []string{"-i", "-F"}
+
+		// We only keep last 10 lines of the stdout and stderr for debugging purposes
+		stdOut := ring.New(10)
```
fgiudici, line 46: nit: since 10 lines is something we put also in the following function, it could be nice to have something like a "DEBUGLINES" const at the beginning of the file

fgiudici (approved):
Looks good here

atanasdinov (approved):
LGTM. An area for exploration would be something like a centralised log parsing output of such complex commands. Perhaps we can extract this outside of the execution itself if we have a lot of different use cases resembling this one.

davidcassany:
> LGTM. An area for exploration would be something like a centralised log parsing output of such complex commands. Perhaps we can extract this outside of the execution itself if we have a lot of different use cases resembling this one.

I am not sure we will have many similar cases, but yes, if we realize there are other places where a similar logging is needed we should extract the logic to a shared location. Usually the problem with this sort of logs handling is that this tends to be different based on the command. In this particular case, the annoying problem is when called inside a container that has not privileges to apply SELinux labels (which is the most common case) then it produces an error line per each file it tries to relabel, basically all the files in the OS 😅

PR #50, opened 2025-04-07
Always run snapper in C locale

dirkmueller:
we're parsing its output, and the output is locale dependant.


davidcassany (approved):
Looks good to me

atanasdinov (approved):
Is this applicable for other commands? Perhaps we could add it as an option if so?

fgiudici (approved):

dirkmueller:
> Is this applicable for other commands? Perhaps we could add it as an option if so?

haven't found one so far. I was thinking about having an option to set the environment in the runner interface. I wouldn't want to hardcode it in general because having localized output of subcommands I think in general is a feature. 

(similarly to elemental tool itself printing localized information)

PR #52, opened 2025-04-09
Include delete excludes in rsync package

davidcassany:
This commit adds additional delete excludes list to protect certain paths of being deleted from destination when mirroring source and destination.

This is mostly helpful to prevent the deletion of mountpoint in target destination that might not be present in source. For instance if an OCI image does not include the /data path but the installation has a partition mounted to /data, in that case we don't want the rsync command to delete /data in destination.


atanasdinov (approved):

fgiudici (approved):
LGTM

pkg/rsync/sync.go
```diff
@@ -74,14 +74,17 @@ func (r Rsync) SyncData(source string, target string, excludes ...string) error
...
+func (r Rsync) MirrorData(source string, target string, excludes []string, deleteExcludes []string) error {
 	flags := r.flags
 	if !slices.Contains(flags, "--delete") {
 		flags = append(flags, "--delete")
 	}
 	for _, e := range excludes {
 		flags = append(flags, fmt.Sprintf("--exclude=%s", e))
 	}
+	for _, e := range deleteExcludes {
+		flags = append(flags, fmt.Sprintf("--filter=protect %s", e))
```
dirkmueller, line 86:
```suggestion
		flags = append(flags, fmt.Sprintf("--filter='protect %s'", e))
```

I'm not sure this actually works (untested) but at least the man page says it should be quoted

davidcassany: I guess man page says that because it assumes to be called within a shell. This is not the case here, the space here is not handled as a separator by any interpreter. In fact I tried with quotes at first and had to remove them. See the same applies for the SLELinux extended attributes filter.

Also, unit tests are actually testing this behavior.

PR #53, opened 2025-04-09
Add btrfs package

davidcassany:
btrfs package is just a wrapper around btrfs tools. Create and delete subvolumes or snapshots and some other actions such as quota handling and default
subvolume setting.

This is essentially required for corner cases where snapper client is not enough or where snapper is not ready yet.

PR #51, opened 2025-04-09
Add fstab package

davidcassany:
This package includes utilities to create and update fstab file.

PR #57, opened 2025-04-10
Add transaction package

davidcassany:
This pull request is actually introducing the transaction package. This package is the responsible of a big chunk of the installation and upgrade of the system. This is the piece of logic that actually handles snapshots, rw volumes and snapshotted volumes as defined in the Deployment struct.

This PR is based on top of #51 and #53

### DIFF
--- a/pkg/transaction/transaction.go
+++ b/pkg/transaction/transaction.go
@@ -43,5 +43,5 @@ type Interface interface {
 	Start(*deployment.ImageSource) (*Transaction, error)
 	Merge(*Transaction) error
 	Commit(*Transaction, Hook, HookBinds) error
-	CloseOnError(*Transaction, error) error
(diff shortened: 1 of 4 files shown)


pkg/transaction/transaction.go
```diff
@@ -0,0 +1,47 @@
...
+	Path       string
+	Merges     map[string]*Merge
+	inProgress bool
+}
+
+type Interface interface {
+	Init(deployment.Deployment) error
+	Start(*deployment.ImageSource) (*Transaction, error)
+	Merge(*Transaction) error
+	Close(*Transaction, Hook, HookBinds) error
```
frelon, line 45: I would prefer to use `Commit` and `Rollback` instead of `Close`/`CloseOnError`

frelon (approved):

PR #58, opened 2025-04-10
Install command

davidcassany:
This PR includes the install package, this is the package that actually handles the full installation of the OS. The PR also includes the execution of the install package from the install command.

Note this PR is based on top of #57


dirkmueller (approved):

fgiudici (approved):
LGTM

PR #64, opened 2025-04-23
Add EFI package

frelon:
The EFI package contains functionality to read/write EFI vars and modify
the boot-entries of the system.

### DIFF
--- a/pkg/efi/manager.go
+++ b/pkg/efi/manager.go
@@ -1,0 +1,198 @@
+// nolint:goheader
+
+// This file is part of nullboot
+// Copyright 2021 Canonical Ltd.
+// SPDX-License-Identifier: GPL-3.0-only
+
+// Package efi contains a boot management library
+package efi
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"math"
+	"path"
+
+	efi "github.com/canonical/go-efilib"
+	efilinux "github.com/canonical/go-efilib/linux"
+
+	"github.com/suse/elemental/v3/pkg/log"
+)
+
+const (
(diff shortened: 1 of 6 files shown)


davidcassany (approved):
LGTM

pkg/efi/manager.go
```diff
@@ -0,0 +1,198 @@
+// nolint:goheader
```
ldevulder, line 1: I supposed that the linter is disabled here because it's not a file created by SUSE for this project, right? In that case maybe just small comment to explain this could be helpful no?

But I agree that with the `Copyright 2021 Canonical Ltd.` below perhaps it's enough?

pkg/efi/manager.go
```diff
@@ -0,0 +1,198 @@
...
+
+// This file is part of nullboot
+// Copyright 2021 Canonical Ltd.
+// SPDX-License-Identifier: GPL-3.0-only
```
ldevulder, line 5: I think we may have an issue here! According to this Apache [document](https://www.apache.org/licenses/GPL-compatibility.html) GPL3 code cannot be included in Apache2.0 one:
```
However, GPLv3 software cannot be included in Apache projects.
The licenses are incompatible in one direction only, and it is a result of ASF's licensing
philosophy and the GPLv3 authors' interpretation of copyright law.
```

ldevulder (changes requested):
I'm a bit worried about the licensing mix, see my comment for more details.

pkg/efi/manager.go
frelon: Good points @ldevulder! I will try to change it! :+1:

PR #66, opened 2025-04-24
Add start sector field in deployment

davidcassany:
This commit adds a startSector attribute in Disk struct. This is used to set the first sector in disk that is used for a partition. This is specially relevant for some ARM devices requiring the ESP partition at a very specific position.

If unset defaults to 1MiB which is 2048 for disks with a sector size of 512 bytes.

### DIFF
--- a/pkg/diskrepart/disk.go
+++ b/pkg/diskrepart/disk.go
@@ -61,28 +61,28 @@ type Disk struct {
 
 type DiskOptions func(d *Disk) error
 
-func WithGdisk() func(d *Disk) error {
+func WithGdisk() DiskOptions {
 	return func(d *Disk) error {
 		d.partBackend = gdiskBack
 		return nil
 	}
 }
 
-func WithParted() func(d *Disk) error {
+func WithParted() DiskOptions {
 	return func(d *Disk) error {
 		d.partBackend = partedBack
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


rdoxenham:
@davidcassany does this block us from setting up large sector size image deployments? e.g. 4096 rather than a standard 512?

davidcassany:
Not really, this should only be used in particular cases in which you know before hand the hardware and parameters. For instance RockChip boards like RK1 require ESP partition to be located at a very specific location. At install time the sector size is detected from the target device, so this defaults to 1MiB and computes the value according to the sector size. If some specific hardware requires a different value for whatever reason this field the chance to configure it, I don't think we should use this without knowing the target hardware constraints.
Again, this is something I needed to build for certain ARM boards, so hardware specific details.

pkg/diskrepart/disk.go
```diff
@@ -81,6 +82,13 @@ func WithBlockDevice(bd block.Device) func(d *Disk) error {
 	}
 }
 
+func WithStartSector(startSector uint) func(d *Disk) error {
```
frelon, line 85: Should this return a `DiskOptions` type?

frelon (approved):

atanasdinov (approved):

PR #67, opened 2025-04-24
Set partition alignement

davidcassany:
Sgdisk uses 1MiB alignement by default on partition start but not on partition end. Partitions partitions are always defined as a multiple of 1MiB with the only exception of the last partition, which takes all available space.

This commit ensures the last partition is also aligned to 1MiB. This turns to be a requirement for encryption tools.


davidcassany:
This prevents this sort of warnings that were appearing on disk verification 

```
Caution: Partition 3 doesn't end on a 2048-sector boundary. This may
result in problems with some disk encryption tools.
```

Apparently having a 1MiB alignment is a strict requirement for LUKS

ldevulder (approved):
LGTM

frelon (approved):

PR #70, opened 2025-04-28
Rename binary to elemental-toolkit

atanasdinov:
- Renames produced binary to elemental-toolkit
- Extracts version command and restructures app creation to allow for reusability
- Next PR will include further separation of the `internal/cli` package: 
  - `internal/cli/elemental/...` 
  - `internal/cli/elemental-toolkit/...`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,5 @@ go.work.sum
 # coverage report
 coverprofile.out
 
-# elemental binary
-elemental
+# elemental binaries
+elemental-toolkit
--- a/cmd/elemental-toolkit/main.go
+++ b/cmd/elemental-toolkit/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/suse/elemental/v3/internal/cli/action"
 	"github.com/suse/elemental/v3/internal/cli/cmd"
+	"github.com/suse/elemental/v3/internal/cli/version"
 	"github.com/urfave/cli/v2"
 )
 
@@ -33,7 +34,7 @@ func main() {
 		cmd.NewInstallCommand(action.Install),
 		cmd.NewUpgradeCommand(action.Upgrade),
 		cmd.NewUnpackImageCommand(action.Unpack),
-		cmd.NewVersionCommand(),
+		version.NewVersionCommand(app.Name),
 	}
 
 	if err := app.Run(os.Args); err != nil {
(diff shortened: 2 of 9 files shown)


cmd/elemental-toolkit/main.go
```diff
@@ -21,23 +21,25 @@ import (
...
-		cmd.NewVersionCommand(),
-	}
-
-	if err := app.Run(os.Args); err != nil {
+	appName := app.Name()
+	application := app.New(
+		cmd.Usage,
+		cmd.GlobalFlags(),
+		cmd.Setup,
+		cmd.NewBuildCommand(appName, action.Build),
```
ipetrov117, line 36: What do we gain from extracting the `appName` as a property of each `Command` function? I understand the general concept behind it, however I wonder if just calling `app.Name()` in the `Command` functions does not offer a cleaner approach.

atanasdinov: While technically possible, this involves additional dependency between the packages which is what I wanted to avoid. See the "graph" below.

`main` -> `action` -> `cmd`
`main` -> `cmd`   (-> `app` if we call `app.Name()` directly)
`main` -> `app`

internal/cli/version/version.go
```diff
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package cmd
+package version
```
fgiudici, line 18: We need to update also the Makefile at lines 12/13, otherwise the version cmd is broken:
```diff
-LDFLAGS+=-X "$(GO_MODULE)/internal/cli/cmd.version=$(GIT_TAG)"
+LDFLAGS+=-X "$(GO_MODULE)/internal/cli/version.version=$(GIT_TAG)"
-LDFLAGS+=-X "$(GO_MODULE)/internal/cli/cmd.gitCommit=$(GIT_COMMIT)"
+LDFLAGS+=-X "$(GO_MODULE)/internal/cli/version.gitCommit=$(GIT_COMMIT)"
```

fgiudici (changes requested):
LGTM, just a missing change required in the Makefile for the version command 👍🏼

.gitignore
```diff
@@ -15,5 +15,5 @@ go.work.sum
 # coverage report
 coverprofile.out
 
-# elemental binary
-elemental
+# elemental binaries
+./elemental-toolkit
```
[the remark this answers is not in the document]
atanasdinov: This is interesting. On my environment it was detecting it as a subdirectory, I'll revert it and adjust in a follow-up if necessary.

fgiudici (approved):

PR #71, opened 2025-04-29
Bootstrap elemental CLI

atanasdinov:
- Adds elemental CLI skeleton and a Makefile target
- Moves the build command from elemental-toolkit to elemental


internal/cli/elemental/cmd/root.go
```diff
@@ -0,0 +1,50 @@
...
+func GlobalFlags() []cli.Flag {
+	return []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "debug",
+			Usage: "Set logging at debug level",
+		},
+	}
+}
+
+func Setup(ctx *cli.Context) error {
```
atanasdinov, line 37: We'll converge the setup mechanism between the different binaries later. We may need similar functionality between the two e.g. the `Runner` interface but the current `System` type includes many utilities that might not be necessary here.

ipetrov117: Agreed. FYI, the release manifest logic will use the `unpack.OCI` functionality (that utilises  the `System` type) for its default way of fetching OCI stored release manifests. That said, this does not mean that we should have the `System` type defined for the whole application by default. The release manifest logic could set it up when needed using `sys.NewSystem()`. I'm adding this mainly for information, I'll rework the code on my side.

frelon (approved):

ipetrov117 (approved):

davidcassany (approved):
LGTM

PR #72, opened 2025-04-30
Support tar sources

davidcassany:
This PR is adding support for Tar sources (.tar, tar.gz and tar.bz2). The motivation is to allow support for the inclusion of arbitrary files as part of the installation. The implementation does not support certain parameters (e.g xattrs and ownership) and file types (e.g. block devices) to keep it simple. It essentially supports directories, files, hardlinks and symlinks and the permissions of them.

Not implemented here yet, but the idea is that the additional files could be a extracted at installation root, so the expectation is that the tarball includes a directory tree based on the OS root. The extraction would be done once the OS snapshot is already read only, hence the tar extraction would only succeed if it includes files to the RW locations (`etc`, `var`, `home`, etc.).


davidcassany:
@atanasdinov I am thinking that `elemental-toolkit install`  customization is defined by three items only:
1. Tarball for extra content (only including files on RW locations otherwise the installation will fail)
1. List of systemd-sysext to install (I believe elemental-toolkit should own systemd-sysext installation and upgrade)
1. Configuration script (I am thinking on making it a parameter, so instead of assuming `config.sh` at a certain path just add flag to set the path to it, I think this will make integration and operation easier)

ipetrov117 (approved):

dirkmueller (approved):

PR #73, opened 2025-04-30
Add GO_EXTRA_ARGS and GO_BUILD_ARGS to makefile

frelon:
These variables can be used to add extra flags to the build-process.

For example the following command should build the go binary in release
mode as done in the uc-factory:

```
make GO_EXTRA_ARGS='-trimpath -buildmode=pie' elemental-toolkit
```

It will also detect if a vendor directory is present and add
'-mod=vendor' to the go build command.


davidcassany (approved):
LGTM

PR #74, opened 2025-05-05
Add overlaytree and config script as installation parameters

davidcassany:
This PR is two fold:

* Refactors transaction interface: the transaction commit method is divided into two: `Commit` and `Close`. Commits sets all required snapshots and its permissions, but it does not set the default snapshot neither runs any sort of customization hook. `Close` method essentially sets the default snapshot and clears/closes all resources associated with the transaction (e.g. mount points). None of the two encompasses any hook execution, hence this refactor is essentially driven to move out the customization logic from the transaction package and simply run it after `Commit`, but before `Close`.

* Adds a configuration script and overlay tree to the deployment. This is essentially to allow setting a custom configuration script and a custom directory tree to overlay over the dumped OS image. This allows extracting the contents of a tarball at OS root and execute a given script just after. These two steps happens between transaction `Commit` and transaction `Close`.

### DIFF
--- a/pkg/install/install_test.go
+++ b/pkg/install/install_test.go
@@ -20,8 +20,6 @@ package install_test
 import (
 	"context"
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 	"testing"
 
@@ -105,14 +103,17 @@ var _ = Describe("Install", Label("install"), func() {
 		runner = sysmock.NewRunner()
 		mounter = sysmock.NewMounter()
 		syscall = &sysmock.Syscall{}
+
 		fs, cleanup, err = sysmock.TestFS(map[string]any{
-			"/etc/config.sh": []byte{},
-			"/dev/device":    []byte{},
-			"/dev/device1":   []byte{},
-			"/dev/device2":   []byte{},
-			"/dev/pts/empty": []byte{},
-			"/proc/empty":    []byte{},
-			"/sys/empty":     []byte{},
+			"/etc/config.sh":  []byte{},
+			"/dev/device":     []byte{},
+			"/dev/device1":    []byte{},
+			"/dev/device2":    []byte{},
+			"/dev/pts/empty":  []byte{},
+			"/proc/empty":     []byte{},
+			"/sys/empty":      []byte{},
+			"/opt/config.sh":  []byte{},
+			"/opt/tree/empty": []byte{},
 		})
 		Expect(err).ToNot(HaveOccurred())
 		s, err = sys.NewSystem(
@@ -126,6 +127,8 @@ var _ = Describe("Install", Label("install"), func() {
 		d.Disks[0].Partitions[0].UUID = "34A8-ABB8"
 		d.Disks[0].Partitions[1].UUID = "34a8abb8-ddb3-48a2-8ecc-2443e92c7510"
 		d.SourceOS = deployment.NewDirSrc("/some/dir")
+		d.CfgScript = "/opt/tree"
+		d.OverlayTree = deployment.NewDirSrc("/opt/tree")
 		Expect(d.Sanitize(s)).To(Succeed())
 		i = install.New(context.Background(), s, install.WithTransaction(t))
 		table = sgdiskEmpty
@@ -155,10 +158,6 @@ var _ = Describe("Install", Label("install"), func() {
 		}
 		t.Trans = trans
 		t.SrcDigest = "imagedigest"
-		dir, err := os.Getwd()
-		Expect(err).ToNot(HaveOccurred())
-		Expect(vfs.MkdirAll(fs, dir, vfs.DirPerm)).To(Succeed())
-		Expect(fs.WriteFile(filepath.Join(dir, "config.sh"), []byte{}, vfs.FilePerm)).To(Succeed())
 	})
 	AfterEach(func() {
 		cleanup()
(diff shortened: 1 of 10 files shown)


pkg/install/install_test.go
```diff
@@ -179,11 +183,30 @@ var _ = Describe("Install", Label("install"), func() {
...
+		Expect(err.Error()).To(ContainSubstring("chroot error"))
+	})
 	It("fails on transaction commit", func() {
 		t.CommitErr = fmt.Errorf("commit failed")
 		err := i.Install(d)
 		Expect(err).To(HaveOccurred())
 		Expect(err.Error()).To(ContainSubstring("commit failed"))
 		Expect(t.RollbackCalled()).To(BeTrue())
 	})
+	It("fails on config hook execution", func() {
```
atanasdinov, line 199: This looks to be a duplicate. Could you please verify this?

pkg/install/install_test.go
```diff
@@ -103,23 +102,33 @@ var _ = Describe("Install", Label("install"), func() {
...
-		Expect(vfs.MkdirAll(fs, "/etc", vfs.DirPerm)).To(Succeed())
-		Expect(fs.WriteFile("/dev/device", []byte{}, vfs.DirPerm)).To(Succeed())
-		Expect(fs.WriteFile("/dev/device1", []byte{}, vfs.DirPerm)).To(Succeed())
-		Expect(fs.WriteFile("/dev/device2", []byte{}, vfs.DirPerm)).To(Succeed())
 		d = deployment.DefaultDeployment()
 		d.Disks[0].Device = "/dev/device"
 		d.Disks[0].Partitions[0].UUID = "34A8-ABB8"
 		d.Disks[0].Partitions[1].UUID = "34a8abb8-ddb3-48a2-8ecc-2443e92c7510"
 		d.SourceOS = deployment.NewDirSrc("/some/dir")
+		d.CfgScript = "/opt/tree"
```
atanasdinov, line 130: Just for the sake of the example, the script should reside outside of the overlay tree, right?

atanasdinov (approved):
This seems like a nice change to unblock the PoC workflow, thanks!

pkg/install/install_test.go
davidcassany: Indeed it is, I am removing it.

pkg/transaction/transaction.go
```diff
@@ -32,16 +37,17 @@ type Merge struct {
...
+	status transactionState
 }
 
 type Interface interface {
 	Init(deployment.Deployment) error
 	Start(*deployment.ImageSource) (*Transaction, error)
 	Merge(*Transaction) error
-	Commit(*Transaction, Hook, HookBinds) error
+	Commit(*Transaction) error
+	Close(*Transaction) error
```
frelon, line 51: Is this interface doing too much? In my eyes a transaction is started, work is done and then committed or rollbacked, this is starting to feel like it is mixing concerns.

Would it make sense to keep the `SetDefault` code in `Commit()` and do the customization between the `Start` and `Commit` calls in `pkg/install/install.go`? Or should SetDefault() be called outside of the transaction?

davidcassany: Another idea I am thinking of is moving some of the logic into the Transaction struct, so splitting it all in two interfaces, however the implementation of both is likely to be coupled, the split would be motivated to identify the duties of each, but not on implementation decoupling, so still sort of a smelly pattern. That could be something like:

```golang
type TransactionManager interface {
  Start(deployment.Deployment) (Transaction, error) // creates a new untouched snapshot and prepares partitions (mounts required volumes) 
  Commit(Transaction) error // sets default snapshot and frees transaction resources
  Rollback(Transaction) error
}

type Transaction interface {
  ID() int                      // retuns transaction ID
  Path() string                 // returns transaction root path
  AddSource(*ImageSource) error // dumps the image source to transaction root and configures snapper
  Merge() error                 // runs 3 way merge logic and updates the fstab file
  Seal() error                  // sets the new snapshot as readonly
}

// Install/Upgrade pseudologic

transaction = TM.Start(d)

transaction.AddSource(OSImage)

transaction.Merge()

// Update deployment file

transaction.Seal()

// Run customization logic here

TM.Commit(transaction)
```

davidcassany: @frelon I finally followed the approach of keeping the same logic in transaction package but organized slightly different across the interface methods. So basically these are the methods that are changed:

`Start`: no longer includes the image source dump and the snapper setup
`Merge`: it doesn't exist anymore, its logic is moved to `Update`.
`Update`: dumps the image source, configures snapper and fstab, runs a setup hook (used for SELinux relabel) and sets the snapshot as read-only.
`Commit`: creates post-transaction snapshots, sets defaults and runs the cleanup logic.

Between `Update` and `Commit` there is a chance to apply the customization logic (extract tarball and run config.sh). Between `Start` and `Update` I don't think we will run any logic, but I think I like this split so `Start` only creates the snapshot path to actually modify and most if not all the changes are all applied on `Update` from an image source.

The logic is all the same, just slightly rearranged. It is hard to try to split it into smaller packages as all the logic around `Start`,  `Update` and  `Commit` is absolutely coupled to our snapper based strategy to handle immutable and RW portions of the system. It is hard to move out logic from transaction package while keeping install and upgrade agnostic to the snapper specific implementation.

frelon: Seems to me like most of the code in Update could be called directly from the installer..


I will approve this for now and we can refine as needed!

PR #78, opened 2025-05-06
Save openQA jobgroup config

ldevulder:
Add a save of the openQA jobgroup.

This file is not saved in openQA configuration, so we have to keep it somewhere. It could be moved elsewhere later (if we have a dedicated repository for QA for example).

### DIFF
--- a/test/cfg/openqa_elemental_jobgroup.yaml
+++ b/test/cfg/openqa_elemental_jobgroup.yaml
@@ -1,0 +1,90 @@
+##########################################################
+#                        WARNING                         #
+#                                                        #
+#              This file is managed in GIT!              #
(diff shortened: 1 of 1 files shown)


test/cfg/openqa_elemental_jobgroup.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+---
+
+.default_products: &default_products
+  distri: sle-micro
+
+.test_settings: &test_settings
+  HDDSIZEGB: '20'
+  QEMURAM: '2048'
+  PASSWORD: ros
```
frelon, line 21: Some things never change 😅

frelon (approved):

fdegir (approved):

PR #79, opened 2025-05-07
Add go.{mod,sum} to GO_FILES

frelon:
This makes the binaries correctly depend on the go.mod and go.sum files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ PKG?=./pkg/... ./internal/...
 
 GO_MODULE ?= $(shell go list -m)
 GO_FILES  = $(shell find ./ -name '*.go' -not -name '*_test.go')
+GO_FILES  += ./go.mod
+GO_FILES  += ./go.sum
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,6 +4,8 @@ PKG?=./pkg/... ./internal/...
 
 GO_MODULE ?= $(shell go list -m)
 GO_FILES  = $(shell find ./ -name '*.go' -not -name '*_test.go')
+GO_FILES  += ./go.mod
+GO_FILES  += ./go.sum
```
ldevulder, line 8: Dumb question maybe, but cannot be written like this?
```
GO_FILES+=./go.mod ./go.sum
```

Seems to work on my lab, but maybe you prefer to keep it in two lines for better readability.

frelon: Yep, it was only a matter of my own personal taste :+1:

ldevulder (approved):

dirkmueller (approved):

PR #80, opened 2025-05-07
Add firmware package

frelon:
The package only contains an EFI boot manager that will call efibootmgr
to create new boot entries.

### DIFF
--- a/pkg/firmware/efi_manager.go
+++ b/pkg/firmware/efi_manager.go
@@ -1,0 +1,86 @@
+/*
+Copyright © 2021 - 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package firmware
+
+import (
+	"path/filepath"
+
+	"github.com/suse/elemental/v3/pkg/sys"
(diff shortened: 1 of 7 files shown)


pkg/firmware/efi_manager.go
```diff
@@ -0,0 +1,86 @@
...
+			b.s.Logger().Error("failed creating boot entry (%s): %s", err.Error(), string(cmdOut))
+			return err
+		}
+	}
+
+	return nil
+}
+
+// DefaultBootEntry generates the default EFI boot entry for the platform.
+func DefaultBootEntry(p *platform.Platform, disk string) *EfiBootEntry {
```
davidcassany, line 70: Would it make sense adding somewhere a sanity check to prevent writing the default boot entry if the target disk is a loop device?
I can't think of any use case where it could be meaningful and it is likely to write on a host that is actually running the installation to craft an image or VM.

I see the default is already false which is good, but I wonder if it wouldn't make sense implementing such an extra sanity check.

davidcassany (approved):
Looks good to me

I think we could consider adding a sanity check to prevent adding the default entry when the target disk is a loop device.

I'd code that in a separate PR though as this might require a small refactor in lsblk wrapper and I am not sure if this is possible for ghw implementation, which I am also considering to entirely drop, lsblk provides more information in a nicely structured way and could drop external go dependencies.

davidcassany:
> This is not very useful yet since we use the fallback boot-entry, but will follow up with some more file installation and actual secure-boot implementation

Note that the vfs package includes the logic to actually find files, when I ported this code form former elemental I was thinking mostly on discovering kernel and initrd files so we can copy them to ESP.
https://github.com/SUSE/elemental/blob/411dd4cbc8d155113a5eaf260cbc8814598b29d9/pkg/sys/vfs/vfs.go#L291-L309

pkg/firmware/efi_manager.go
davidcassany: Yes that's true, fair enough. We should be careful not to document anywhere a potentially harmful receipt to create an image.

ipetrov117 (approved):

PR #84, opened 2025-05-09
Switch naming to elemental3 prefix to avoid confusion with 1.x

dirkmueller:


frelon (approved):

ldevulder (commented):
I think it's better to have it in the RPM spec file rather than here, as the same repo can contains a v4 version later.

dirkmueller:
> I think it's better to have it in the RPM spec file rather than here, as the same repo can contains a v4 version later.

my concern there is that devs don't test then what the users (via rpm) will get.

davidcassany (approved):

PR #85, opened 2025-05-09
Install bootloader

frelon:
This PR adds the bootloader package which is responsible for installing the bootloader.

The 'none' bootloader does nothing and leaves the bootloader implementation to the OCI image. (anything in /boot/efi will be copied to the ESP and used as a bootloader).

The default 'grub' bootloader install grub and tries to emulate the BLS specification by placing grub-env files in `$ESP/grubenv` and `$ESP/loader/entries`. 
It will also install shim/mokmanager/grub.efi into `/EFI/ELEMENTAL` and the fallback `/EFI/BOOT` directories and sync `/usr/share/grub2` to `$ESP/grub2`.
After all that it will copy the kernel and generate an initrd in `$ESP/<os-variant>/<kernel-version>/`

### DIFF
--- a/pkg/bootloader/grub.go
+++ b/pkg/bootloader/grub.go
@@ -47,13 +47,13 @@ func NewGrub(s *sys.System) *Grub {
 }
 
 const (
-	OsReleasePath  = "/etc/os-release"
-	DefaultCmdline = "rw quiet"
-	Initrd         = "initrd"
+	OsReleasePath = "/etc/os-release"
+	Initrd        = "initrd"
 )
 
 // Install installs the bootloader to the specified root.
-func (g *Grub) Install(rootPath string, esp *deployment.Partition) error {
+func (g *Grub) Install(rootPath string, d *deployment.Deployment) error {
+	esp := d.GetEfiSystemPartition()
 	if esp == nil {
 		g.s.Logger().Error("ESP not found")
 		return fmt.Errorf("ESP not found")
@@ -81,6 +81,10 @@ func (g *Grub) Install(rootPath string, esp *deployment.Partition) error {
 	entry, err := g.installKernelInitrd(rootPath, esp)
 	if err != nil {
 		g.s.Logger().Error("Error installing kernel+initrd: %s", err.Error())
+	}
+
+	if d.BootConfig != nil {
+		entry.cmdline = d.BootConfig.KernelCmdline
 	}
 
 	return g.updateBootEntries(rootPath, esp, entry)
@@ -197,7 +201,6 @@ func (g *Grub) installKernelInitrd(rootPath string, esp *deployment.Partition) (
 		linux:       filepath.Join("/", osID, kernelVersion, filepath.Base(kernel)),
 		initrd:      filepath.Join("/", osID, kernelVersion, Initrd),
 		displayName: displayName,
-		cmdline:     DefaultCmdline,
 		id:          "active",
 	}, nil
 }
--- a/pkg/bootloader/grub.go
+++ b/pkg/bootloader/grub.go
@@ -1,0 +1,40 @@
+/*
+Copyright © 2022-2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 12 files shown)


pkg/install/install.go
```diff
@@ -111,6 +122,14 @@ func (i Installer) Install(d *deployment.Deployment) (err error) {
 		return err
 	}
 
+	err = i.b.Install(trans.Path, d)
```
davidcassany, line 125: I think this should be replaced after the config script hook, this way if some additional config (e.g. a new initrd) is applied the bootloader installer can actually pick the updated data.

davidcassany (changes requested):
Nice! I like it.

@frelon I noted some details which are not really affecting the core logic of the PR, I am open for discussion if you have a different view

frelon:
> Nice! I like it.
> 
> 
> 
> @frelon I noted some details which are not really affecting the core logic of the PR, I am open for discussion if you have a different view

Great, I will make some updates based on your comments! 👍

pkg/install/install.go
```diff
@@ -111,6 +122,14 @@ func (i Installer) Install(d *deployment.Deployment) (err error) {
...
+	if err != nil {
+		i.s.Logger().Error("installation failed, could not install bootloader: %s", err.Error())
+		return err
+	}
+
+	err = d.WriteDeploymentFile(i.s, trans.Path)
```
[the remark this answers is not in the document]
frelon: Yep, think I messed up resolving conflict, will fix!

pkg/bootloader/grub.go
```diff
@@ -0,0 +1,247 @@
...
+	if exists, _ := vfs.Exists(g.s.FS(), expectedInitrdPath); exists {
+		// Initrd exists, copy it.
+		err = vfs.CopyFile(g.s.FS(), expectedInitrdPath, targetDir)
+		if err != nil {
+			g.s.Logger().Info("Error copying initrd '%s': %s", targetDir, err.Error())
+			return grubBootEntry{}, err
+		}
+	} else {
+		// Initrd does not exist, generate a new one
+		initrdPath := filepath.Join(esp.MountPoint, osID, kernelVersion, Initrd)
```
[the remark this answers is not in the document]
frelon: Good point, will remove this functionality!

pkg/install/install.go
frelon: Interesting, I put this before the config-script to be able to make changes to the bootloader setup, but I don't have strong feelings about this so I will move it to after the hook!

pkg/sys/vfs/vfs.go
```diff
@@ -564,3 +565,89 @@ func ConcatFiles(fs FS, sources []string, target string) (err error) {
 
 	return fs.Chmod(target, fInf.Mode())
 }
+
+// LoadEnvFile will try to parse the file given and return a map with the key/values
+func LoadEnvFile(fs FS, file string) (map[string]string, error) {
```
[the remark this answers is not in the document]
frelon: Added tests.

davidcassany (approved):
LGTM :+1:

Note: I am not sure if during the installa some of the data copied to ESP partitions is already assumed to be in ESP mountpoint, but if this is the case we need to consider special behavior during upgrades. On upgrades the image contents which are part of the RW volumes is not synched (there shouldn't be anything in there, a directory structure at most).

pkg/bootloader/grub.go
```diff
@@ -0,0 +1,233 @@
...
+		return grubBootEntry{}, fmt.Errorf("initrd not found")
+	}
+
+	err = vfs.CopyFile(g.s.FS(), expectedInitrdPath, targetDir)
+	if err != nil {
+		g.s.Logger().Info("Error copying initrd '%s': %s", targetDir, err.Error())
+		return grubBootEntry{}, err
+	}
+
+	displayName, ok := osVars["PRETTY_NAME"]
```
dirkmueller, line 192: starting with SLES16 we actually have the displayName in `VARIANT` instead of PRETTY_NAME. can we handle that here?

frelon: Thanks for the info! I added a check so the order goes PRETTY_NAME > VARIANT > NAME.

dirkmueller (approved):

PR #86, opened 2025-05-12
Refactor elemental install command line flags

davidcassany:
This commit introduces the overlay and description flags. The overlay flag is to point to a source to overlay on top of the deployed OS image (e.g. a tarball containing some extra utilities that are not part of the OS image). Note the overlayed content is already included once the OS snapshot is defined as read-only, hence it is only possible to overlay on top of RW portions of the system.

The description flag is what the former config flag used to be. It is used to set the local path of the deployment description yaml file.

Finally the former config flag is repurposed to be set the local path of the configuration script that is called in chroot during the transaction. It is executed just after extracting the overlay content, if any. So the script is only expected to modify RW parts of the deployment.

### DIFF
--- a/internal/cli/elemental-toolkit/action/install.go
+++ b/internal/cli/elemental-toolkit/action/install.go
@@ -72,13 +72,13 @@ func Install(ctx *cli.Context) error { //nolint:dupl
 
 func digestInstallSetup(s *sys.System, flags *cmd.InstallFlags) (*deployment.Deployment, error) {
 	d := deployment.DefaultDeployment()
-	if flags.ConfigFile != "" {
-		if ok, _ := vfs.Exists(s.FS(), flags.ConfigFile); !ok {
-			return nil, fmt.Errorf("config file '%s' not found", flags.ConfigFile)
+	if flags.Description != "" {
+		if ok, _ := vfs.Exists(s.FS(), flags.Description); !ok {
+			return nil, fmt.Errorf("config file '%s' not found", flags.Description)
 		}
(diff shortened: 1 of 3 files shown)


ldevulder (approved):
LGTM!

For now the openQA test doesn't use `config` option, so that's fine, but I'll keep this in mind.

ipetrov117 (approved):

frelon (commented):
s/elemtal/elemental in the commit-message and PR header :+1:

internal/cli/elemental-toolkit/action/install.go
```diff
@@ -72,13 +72,13 @@ func Install(ctx *cli.Context) error { //nolint:dupl
...
-			return nil, fmt.Errorf("config file '%s' not found", flags.ConfigFile)
+	if flags.Description != "" {
+		if ok, _ := vfs.Exists(s.FS(), flags.Description); !ok {
+			return nil, fmt.Errorf("config file '%s' not found", flags.Description)
 		}
-		data, err := s.FS().ReadFile(flags.ConfigFile)
+		data, err := s.FS().ReadFile(flags.Description)
 		if err != nil {
-			return nil, fmt.Errorf("could not read config file '%s': %w", flags.ConfigFile, err)
+			return nil, fmt.Errorf("could not read config file '%s': %w", flags.Description, err)
```
dirkmueller, line 81:
```suggestion
			return nil, fmt.Errorf("could not read description file '%s': %w", flags.Description, err)
```

PR #89, opened 2025-05-12
Introduce release manifest resolution logic

ipetrov117:
This commit introduces the release manifest resolution logic that will be used by the `elemental` tool during the `elemental build` command. To reduce the size of this PR, unit tests will come in a separate PR once this one has been reviewed and merged.

Review entry point - [resolver.go](https://github.com/SUSE/elemental/pull/89/files#diff-97772d0b7c29df5038c7d3233a80ecad7b5827ca88c4ac2da678cd9e3d5dc753)

Changes from the `Release Manifest Design Doc`:

- `basePlatform` configuration in `product` manifest was renamed to `corePlatform` for consistency.

Key points of the release manifest resolution logic:

1. Release manifest is parsed to a `ResolvedReleaseManifest` struct that holds both `product` and `core` release manifest data that will be used to retrieve the necessary data during `elemental build`.
2. Can parse both `core` and `product` release manifests.
3. Release manifest sources are provided in URI format. Currently `oci` and `file` are supported:
    * `oci://registry.com/foo/bar/release-manifest:0.0.1`
    * `fiile:///foo/bar/release-manifest.yaml`
4. `Core` manifests are referred only by version in the `product` release manifests - the construction of the valid OCI image for the `core` manifest is done in-code. This limits the chance for a user to provide an OCI image or file that points to a not supported/tested set of core component versions. For testing purposes, the default value of the `core` manifest registry will be configurable during the build step. The specifics of this will be handled during the integration of this logic and the `build` command.

Manifest resolution process:
```
-> Parse URI to 'ReleaseManifestSource'
-> Read manifest data from 'ReleaseManifestSource'
  -> [OCI source] -> Unpack OCI FS -> Locate manifest -> Move manifest to store (can be later used as cache) -> Remove OCI FS -> Read manifest from store
-> Try to parse data:
  -> To 'product' manifest -> To 'core' manifest -> Fail
-> [product manifest] Construct `core` manifest OCI image and repeat resolution 
-> [core manifest] Return the resolved manifest in the form of 'ResolvedReleaseManifest' struct.
```

Enhacements post-POC:

1. Validation of manifest API
2. OCI caching mechanism based on the `store` logic introduced in the `OCIReleaseManifestExtractor`
3. Local OCI image handling

### DIFF
--- a/internal/manifest/extractor/extractor.go
+++ b/internal/manifest/extractor/extractor.go
@@ -1,0 +1,221 @@
+/*
+Copyright © 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extractor
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/suse/elemental/v3/pkg/sys"
+	"github.com/suse/elemental/v3/pkg/unpack"
+)
+
+// Default glob pattern used to search for a release manifest in an OCI image
+const globPattern = "release_manifest*.yaml"
+
+type OCIUnpacker interface {
+	// Unpack unpacks the file system of a given OCI image to the specified destination
+	// and returns its digest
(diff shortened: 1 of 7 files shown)


davidcassany (approved):
Looks good to me :+1:

I see the packages to abstract from the host (pkg/vfs and anything under pkg/sys) are not used from the system host, no strong opinion, but probably we should standardize at some point.

ipetrov117:
>I see the packages to abstract from the host (pkg/vfs and anything under pkg/sys) are not used from the system host, no strong opinion, but probably we should standardize at some point.

I agree on the standardisation point, my main reasoning for not using `pkg/sus` or `pkg/vfs` is that I am not 100% convinced that the release manifest library will reside in this repository in the long run, so I opted to have something more generic. Once we have better understanding on the usage of this library, we can standardise it, if it is okay with you.

davidcassany:
> I agree on the standardisation point, my main reasoning for not using pkg/sys or pkg/vfs is that I am not 100% convinced that the release manifest library will reside in this repository in the long run

Indeed, I fully agree. Let's get a wider picture first

frelon (approved):

PR #91, opened 2025-05-15
Introduce release manifest unit tests

ipetrov117:
This commit introduces test for the release manifest resolution logic that came with #89.

### DIFF
--- a/internal/manifest/extractor/extractor_test.go
+++ b/internal/manifest/extractor/extractor_test.go
@@ -1,0 +1,212 @@
+/*
+Copyright © 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package extractor_test
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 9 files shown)


ldevulder (approved):
Thanks for taking care of the unit tests! LGTM!

internal/manifest/extractor/extractor_test.go
```diff
@@ -0,0 +1,212 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Release Manifest OCI Extractor test suite")
+}
+
+var _ = Describe("OCIReleaseManifestExtractor", Label("release-manifest"), func() {
+	var customStoreRoot string
+	var unpacker *unpackerMock
+	BeforeEach(func() {
+		var err error
+		customStoreRoot, err = os.MkdirTemp("", "extractor-custom-store-")
```
frelon, line 52: I would consider using the vfs mock to handle filesystem access to be consistent with other tests (see for example https://github.com/SUSE/elemental/blob/67cdffacf8f3eb64fc88cbea4ec268eaa3c7506e/pkg/unpack/oci_test.go#L45)

frelon (approved):

internal/manifest/extractor/extractor_test.go
ipetrov117: I considered this, but opted against it. Mainly because as of this moment, `extractor.go` is not bound to the `vfs.FS` filesystem.  My main reason for not using `vfs.FS` in `extractor.go` was that I was (and still am) unsure where the release manifest resolution logic will end up in the long run. But taking into account that the `extractor.go` is under `internal`, it might be a good idea for it to use `vfs.FS`, at least until we decide on the correct placement of this logic.

I will merge this as it is and open a follow-up PR, which introduces the `vfs.FS` usage in the `extractor.go` and `extractor_test.go` files. Sound good?

frelon: Sure, nice work! 👍

PR #93, opened 2025-05-16
Make unpack configurable

davidcassany:
This commit makes unpacker factory configurable. It maps the configuration options of the underlaying implementations to the factory method.

The end goal of this change is to facilitate mapping OCI extractor options such as the local flag to upper layers in the code such as the command line interface.

In addition it also allows to set a different set of rsync flags during the overlay tree extraction which is required in order to use dir:// sources.


davidcassany:
@ipetrov117 with this change it will be trivial to workaround the issue you found in setting the overlay as a `dir://` source.

frelon (approved):

dirkmueller (approved):

PR #98, opened 2025-05-23
Upgrade refactor

davidcassany:
This PR is a follow up from https://github.com/SUSE/elemental/pull/74#issuecomment-2865752659 , basically removes the `Update` method from the transaction interface and breaks it into smaller pieces as part of a new `UpgradeHelper` interface. Basically the upgrade helper interface implements the upgrades steps that are coupled to the transaction implementation details (snapper snapshots in this particular case).

Since both interfaces are coupled to its particular implementation an UpgradeHelper instance is returned as part of the `Init` method of the transaction interface. This way it is the transaction implementation who provides a matching helper instance.

Both interfaces implementations are kept together into the same package as they are strictly coupled, in fact, both implementations use the same underlying struct to track internal metadata.

In addition this also collects the upgrade steps into an `Upgrader` object which is is called as part of the upgrade and install processes. So that installation partitions the disk and after that _upgrades_ to the required image.

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -23,47 +23,26 @@ import (
 
 	"github.com/suse/elemental/v3/pkg/block"
 	"github.com/suse/elemental/v3/pkg/block/lsblk"
-	"github.com/suse/elemental/v3/pkg/bootloader"
 	"github.com/suse/elemental/v3/pkg/btrfs"
-	"github.com/suse/elemental/v3/pkg/chroot"
 	"github.com/suse/elemental/v3/pkg/cleanstack"
 	"github.com/suse/elemental/v3/pkg/deployment"
 	"github.com/suse/elemental/v3/pkg/diskrepart"
-	"github.com/suse/elemental/v3/pkg/firmware"
-	"github.com/suse/elemental/v3/pkg/selinux"
 	"github.com/suse/elemental/v3/pkg/sys"
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
-	"github.com/suse/elemental/v3/pkg/transaction"
-	"github.com/suse/elemental/v3/pkg/unpack"
+	"github.com/suse/elemental/v3/pkg/upgrade"
 )
-
-const configFile = "config.sh"
 
 type Option func(*Installer)
 
 type Installer struct {
 	ctx context.Context
 	s   *sys.System
-	t   transaction.Interface
-	bm  *firmware.EfiBootManager
-	b   bootloader.Bootloader
+	u   upgrade.Interface
 }
 
-func WithTransaction(t transaction.Interface) Option {
+func WithUpgrader(u upgrade.Interface) Option {
 	return func(i *Installer) {
-		i.t = t
-	}
-}
-
-func WithBootManager(bm *firmware.EfiBootManager) Option {
-	return func(i *Installer) {
-		i.bm = bm
-	}
-}
-
-func WithBootloader(b bootloader.Bootloader) Option {
-	return func(i *Installer) {
-		i.b = b
+		i.u = u
 	}
 }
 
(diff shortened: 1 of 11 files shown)


pkg/install/install.go
```diff
@@ -103,55 +79,9 @@ func (i Installer) Install(d *deployment.Deployment) (err error) {
 		}
 	}
 
-	err = i.t.Init(*d)
+	err = i.u.Upgrade(d)
```
davidcassany, line 82: Install is a particular case of upgrade with this change. The upgrade procedure does not require a previous snapshot to be present.

pkg/transaction/snapper_update_helper.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/suse/elemental/v3/pkg/chroot"
+	"github.com/suse/elemental/v3/pkg/deployment"
+	"github.com/suse/elemental/v3/pkg/fstab"
+	"github.com/suse/elemental/v3/pkg/rsync"
+	"github.com/suse/elemental/v3/pkg/snapper"
+	"github.com/suse/elemental/v3/pkg/sys/vfs"
+	"github.com/suse/elemental/v3/pkg/unpack"
+)
+
+func (sc snapperContext) SyncImageContent(imgSrc *deployment.ImageSource, trans *Transaction, fullSync bool) (err error) {
```
davidcassany, line 34: The methods of this file are essentially the former `transaction.Update` method splitting each step in an independent method. This allows further granularity to execute additional logic between those and it removes the need of the upgrade callbacks.

pkg/transaction/transaction.go
```diff
@@ -42,13 +41,20 @@ type Transaction struct {
 	ID     int
 	Path   string
 	Merges map[string]*Merge
+
 	status transactionState
 }
 
 type Interface interface {
-	Init(deployment.Deployment) error
+	Init(deployment.Deployment) (UpgradeHelper, error)
```
davidcassany, line 49: This makes explicit the coupling between the UpgradeHelper and Transaction interfaces.

pkg/transaction/snapper.go
```diff
@@ -44,45 +40,67 @@ const (
...
 	ctx          context.Context
 	s            *sys.System
+	partitions   deployment.Partitions
+	cleanStack   *cleanstack.CleanStack
 	snap         *snapper.Snapper
+	maxSnapshots int
+}
+
+// checkCancelled just checks if the current context is cancelled or not. Returns the context error if cancelled.
+func (sc snapperContext) checkCancelled(err error) error {
```
frelon, line 53: Is this needed? Can this be 

```
if err2 := ctx.Err(); err2 != nil {
    err = err2
}
```

In the deferred calls? (I have not tried this myself yet, channels are always a bit of magic)

pkg/upgrade/upgrade.go
```diff
@@ -19,87 +19,188 @@ package upgrade
...
+
+	err = d.WriteDeploymentFile(u.s, trans.Path)
+	if err != nil {
+		u.s.Logger().Error("could not write deployment file")
+		return err
+	}
+
+	err = uh.Lock(trans)
+	if err != nil {
+		u.s.Logger().Error("failed relabelling snapshot path: %s", trans.Path)
```
frelon, line 137: Seems like this error message was copy-pasted from above?

fgiudici (approved):
Looks good here!

pkg/transaction/snapper_update_helper.go
```diff
@@ -0,0 +1,303 @@
+/*
```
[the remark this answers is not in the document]
davidcassany: Ok I split the tests, now there snapper_test.go testing snapper.go code and snapper_upgrade_helper_test.go testing snapper_upgrade_helper.go code. I guess this way is way more explicit. Since some logic is shared between the two I moved that part in transaction_test.go and it is referenced in both test files.

Unfortunately I had to rebase the whole PR not to conflic with snapper_test.go changes from #99

dirkmueller (approved):

PR #102, opened 2025-06-02
Set relaxed rsync flags for overlay tree sync

davidcassany:
This PR is to enable using `dir://` sources for the `--overlay` flag in install procedures. It basically follows a similar criteria as it is done for tarball sources. Original ownership is not kept and in case of overwrites it does not change pre-existing permissions in target destination.


davidcassany:
@ipetrov117 with this you should be capable to use `dir://` as paths as the value for the `--overlay` flag, which is handy for quick iterations.

ipetrov117 (approved):
LGTM!

PR #103, opened 2025-06-02
Set default ESP mount-opts to x-systemd.automount

frelon:
This adds an automount unit for the /boot mountpoint which will mount
the ESP when used.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -368,6 +368,7 @@ func DefaultDeployment() *Deployment {
 					MountPoint: EfiMnt,
 					FileSystem: VFat,
 					Size:       EfiSize,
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
Is this to avoid having the /boot partition mounted at boot by default? If I get it right this will be only mounted if something tries to access /boot.

frelon:
> Is this to avoid having the /boot partition mounted at boot by default? If I get it right this will be only mounted if something tries to access /boot.

Exactly, systemd-automount will only mount /boot when something tries to access it, and also parallelize the mounting during boot.

I did not see any boot time improvement, but unmounting /boot and then trying to access it remounted it, which is definitely nice if /boot is ever unmounted and we then try to run an upgrade.

PR #104, opened 2025-06-03
Add config and overlay flags for update

davidcassany:
This commit adds the --overlay and --config flags to upgrade subcommand. They behave exactly in the same way as they do in install subcommand.

### DIFF
--- a/internal/cli/elemental-toolkit/cmd/upgrade.go
+++ b/internal/cli/elemental-toolkit/cmd/upgrade.go
@@ -40,18 +40,18 @@ func NewUpgradeCommand(appName string, action func(*cli.Context) error) *cli.Com
 		Flags: []cli.Flag{
 			&cli.StringFlag{
 				Name:        "os-image",
-				Usage:       "image containing the operating system",
+				Usage:       "URI to the image containing the operating system",
 				Destination: &UpgradeArgs.OperatingSystemImage,
 				Required:    true,
 			},
 			&cli.StringFlag{
 				Name:        "config",
-				Usage:       "Configuration script path to execute after committing os-image",
+				Usage:       "Path to OS image post-commit script",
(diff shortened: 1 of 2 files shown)


davidcassany:
I think for now this will be helpful for the time being, specially if we consider an upgrade in which extensions (e.g. toolbox) and some configurations are needed.

frelon (approved):

internal/cli/elemental-toolkit/cmd/upgrade.go
```diff
@@ -42,6 +44,16 @@ func NewUpgradeCommand(appName string, action func(*cli.Context) error) *cli.Com
...
 				Required:    true,
 			},
+			&cli.StringFlag{
+				Name:        "config",
+				Usage:       "Configuration script path to execute after committing os-image",
+				Destination: &UpgradeArgs.ConfigScript,
+			},
+			&cli.StringFlag{
+				Name:        "overlay",
+				Usage:       "URI to the content overlaid over the OS image",
```
dirkmueller, line 54:
```suggestion
				Usage:       "URI of the overlay content for the OS image",
```

internal/cli/elemental-toolkit/cmd/upgrade.go
```diff
@@ -42,6 +44,16 @@ func NewUpgradeCommand(appName string, action func(*cli.Context) error) *cli.Com
 				Destination: &UpgradeArgs.OperatingSystemImage,
 				Required:    true,
 			},
+			&cli.StringFlag{
+				Name:        "config",
+				Usage:       "Configuration script path to execute after committing os-image",
```
dirkmueller, line 49:
```suggestion
				Usage:       "Path to OS image post-commit script",
```

maybe?

PR #109, opened 2025-06-10
Improve error propagation

atanasdinov:
Error propagation always occurs and ends at the top of the chain right where the CLI command execution finishes. If error is present, it will be logged. This is why it's very important to carry on context for each error e.g. opening file error without any wrapping would be extremely hard to identify.

This PR drops a lot of repetitive logging, improves error reporting, adds minor fixes in the process and adjusts tests accordingly.

Some of the error values now do include command outputs which in the future would be extracted to the proper log files instead.


frelon:
FWIW I disagree with this approach and much prefer to log errors closer to the source. Trying to append all context to an error and returning it instead of having a clear log in my opinion is a worse developer experience.

PR #110, opened 2025-06-11
Grub fix boot entry on install

frelon:
The 'active' boot entry was appended on upgrades, leaving the
/boot/grubenv entries=active 1 active 2.

This commit skips adding the second 'active' and sorts the list
correctly, leaving us with:
/boot/grubenv entries=active 2 1


pkg/bootloader/grub.go
```diff
@@ -24,6 +24,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"golang.org/x/exp/slices"
```
davidcassany, line 27: Wasn't it enough with "slices"? I think Reverse is already part of the [standard library](https://pkg.go.dev/slices#Reverse)

davidcassany (commented):
Looks good to me, I just have doubts about requiring `golang.org/x/exp`, unless I missed something to me it looks like there is no special need for that, we already have the required functionality in `slices` from standard library.

pkg/bootloader/grub.go
```diff
@@ -248,19 +250,22 @@ func (g *Grub) updateBootEntries(rootPath string, esp *deployment.Partition, new
...
 	// Read current entries
-	stdOut, err := g.s.Runner().Run("grub2-editenv", grubEnvPath, "list")
-	if err != nil {
-		return fmt.Errorf("reading current boot entries: %w", err)
-	}
+	if ok, _ := vfs.Exists(g.s.FS(), grubEnvPath); ok {
+		stdOut, err := g.s.Runner().Run("grub2-editenv", grubEnvPath, "list")
+		if err != nil {
+			g.s.Logger().Error("Failed reading current boot entries: %s", err.Error())
+			return err
```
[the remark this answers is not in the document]
frelon: Sure. Although next time we do changes that affect the error handling of the entire application it would be good to get buy-in from the whole team.

davidcassany (approved):

pkg/bootloader/grub.go
```diff
@@ -248,19 +249,21 @@ func (g *Grub) updateBootEntries(rootPath string, esp *deployment.Partition, new
...
+			return fmt.Errorf("reading current boot entries: %w", err)
+		}
 
-	g.s.Logger().Debug("grub2-editenv stdout: %s", string(stdOut))
-	for line := range strings.SplitSeq(string(stdOut), "\n") {
-		if after, found := strings.CutPrefix(line, "entries="); found {
-			activeEntries = append(activeEntries, strings.Fields(after)...)
+		g.s.Logger().Debug("grub2-editenv stdout: %s", string(stdOut))
+		for line := range strings.SplitSeq(string(stdOut), "\n") {
+			if after, found := strings.CutPrefix(line, "entries=active "); found {
```
ldevulder, line 260: Couldn't be a problem later to have `active` hardcoded here? As there is a constant for this maybe better to use `strings.CutPrefix(line, "entries="+DefaultBootID+" ")`?

PR #111, opened 2025-06-12
Simple 3 way merge for upgrades

davidcassany:
This commit implements a simple 3 way merge for the snapshotted rw volumes (like it happens for /etc). The current implementation feeds the volume with the contents of the new image and on top of that it applies any modified file comparing former image content versus the current content before the upgrade.

The implementation is based on snapper status reports to list modified and deleted files. Any files listed as a deletion is deleted in target and any modified file is rsynced to the target. So local modifications of mutable files will always "win" vs modifications provided by an upgrade image.

For instance, when you change the contents of /etc in a host, any added, modified or removed files will be applied to the new snapshot on upgrades.

### DIFF
--- a/pkg/transaction/snapper_upgrade_helper.go
+++ b/pkg/transaction/snapper_upgrade_helper.go
@@ -18,8 +18,13 @@ limitations under the License.
 package transaction
 
 import (
+	"bufio"
 	"fmt"
+	"os"
 	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
 
 	"github.com/suse/elemental/v3/pkg/btrfs"
 	"github.com/suse/elemental/v3/pkg/chroot"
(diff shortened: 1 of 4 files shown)


pkg/transaction/snapper_upgrade_helper.go
```diff
@@ -207,22 +212,162 @@ func (sc snapperContext) configureRWVolumes(trans *Transaction) error {
...
+		switch {
+		case len(match) == 0:
+			continue
+		case match[1] == "....":
+			// Ignore extended attributes changes because the stock snapshot used for
+			// comparison was taken before SELINUX relabelling, hence this is likely to
+			// list almost every single file.
+			continue
+		case match[2] == "-":
+			err = sc.s.FS().RemoveAll(filepath.Join(merge.New, match[3]))
```
dirkmueller, line 314: do I get this correctly: so we delete the file even if it was modified locally on the system? how about we rename it to a backup file?

pkg/transaction/snapper_upgrade_helper.go
```diff
@@ -207,22 +212,162 @@ func (sc snapperContext) configureRWVolumes(trans *Transaction) error {
...
+		line := scanner.Text()
+		match := r.FindStringSubmatch(line)
+
+		switch {
+		case len(match) == 0:
+			continue
+		case match[1] == "....":
+			// Ignore extended attributes changes because the stock snapshot used for
+			// comparison was taken before SELINUX relabelling, hence this is likely to
+			// list almost every single file.
```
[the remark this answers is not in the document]
davidcassany: how to handle SELinux labels is something to improve I believe. in rsync calls we are omitting selinux labels by default, but no other extended attributes. The OCI image tree is not supposed to include selinux attributes (selinux labels belong to the host not to the container image in a regular containers runtime context).

So yes, I think this needs to be improved, but this requires a closer look on how to handle SELinux labels in general. The current approach just applies labels once to the whole tree after the sync and before locking immutable volumes, I guess this is just a too simplistic approach. Probably we should apply them before making any /etc snapshot and then reapply them after customization took place.

IMHO this is belonging to a SELinux related effort though, not limited to the merging strategy.

pkg/transaction/snapper_upgrade_helper.go
davidcassany: This is to honor the following use case.

* Former installed image included `/etc/somefile`
* As part of a customization process `/etc/somefile` got deleted
* OS image to upgrade to contains `/etc/somefile` (modified or not), but since the user deleted it in the past, we delete it again.

So the suggested merge approach in this PR, has a strict priority of customizations vs changes provided through images.

The upgraded `/etc/somefile` in the new image is already kept in `stock=true` snapshot of `/etc`. So the content is not lost and we can still diff it. We could argue if we make a backup in the current volume too, but I am not sure this actually makes sense unless we have a process to deal with it, so I'd not backup/copy of it for the time being.

Another _feature_ that could be implemented here is making a status diff between the old image and new image, so we could also list the files changed between the two images and check if there is any conflict (at least drop a warning for the time being). This was the next step I kind a have in mind.

pkg/transaction/snapper_upgrade_helper.go
```diff
@@ -207,22 +212,162 @@ func (sc snapperContext) configureRWVolumes(trans *Transaction) error {
...
+	err = sc.snap.Status(root, snapper.ConfigName(volPath), output, oldID, modifiedID)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// applyCustomChanges reads the given status file and applies reported changes in to the target destination.
+// This method is the responsible of applying customizations to the new volume
```
ipetrov117, line 278: nit:
```suggestion
// This method is responsible for applying customizations to the new volume
```

ipetrov117 (approved):
LGTM

pkg/transaction/snapper_upgrade_helper.go
dirkmueller: I see. I agree this is what the logic should become:
* no change between oldstock and newstock: no change (no delete, no overwrite)
* was there in oldstock but is not in newstock: delete with backup? delete always?
* was not in oldstck but is in newstock: add
* was in old and new stock with a difference: ? overwrite? delete? custom logic lookup to deal with it case by case? 

this needs to become part of the release manifest, where we can decide what to do with modifications (preserve, overwrite, apply)

davidcassany: Yes, this is interesting, what I had in mind is that non conflicting changes are always applied no matter if defined by local modifications or defined by a new stock. This is what I imagined, but yes probably we can make this configurable too.

Then in case of conflicting changes (new stock introduces changes to a file and local modification also introduced changes to the same file) I imagined three different scenarios:

1. conflict is solved by ignoring new stock changes and applying local changes (this is what this PR implements)
2. conflict is solved by ignoring local changes and applying new stock changes only (I struggle to see it something we want)
3. there is some _smart_ logic to adjust whatever is needed to _keep_ both modifications (this is implies providing a similar concept to RPM scriptlets as I can't see any way to generalize it)

The problem of this approach is if single change encompasses multiple files (e.g. a config file split into several files), then approach 1. and 2. would produce an inconsistent result in any case. So considering a process to provide case by case logic is something we might have to consider. The challenge there is having a process in place to monitor changes across image rebuilds and upgrades to automate at the very least the potentially problematic cases so before releasing a new image we can elaborate a some custom logic to run in case there is a conflict. Feels convoluted, but probably it turns out there aren't that many cases to consider. I'd say the key is monitoring/comparing builds before releasing to detect changes in /etc.

PR #113, opened 2025-06-13
Allow enabling Helm chart extensions from product release manifest

atanasdinov:
Initial implementation of Helm chart enablement.

The source is always the **product release manifest** for now. Further development will tell us whether we allow the **core release manifest** to also be leveraged for this functionality, for example enabling MetalLB for single node use cases.

Additionally, we might want to improve the UX in the future via the release manifest API. The current implementation allows the users to enable dependencies only (e.g. `neuvector-crd`) which isn't really useful on its own. This warrants further discussion.

Also extracts out the current Kubernetes and Helm related functionality to separate files within the `build` package as `build.go` was starting to grow too much.

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -60,13 +60,13 @@ func Run(ctx context.Context, d *image.Definition, buildDir string, system *sys.
 		return err
 	}
 
-	if len(d.Release.Enable) != 0 {
-		logger.Info("Enabling the following product extensions: %s", strings.Join(d.Release.Enable, ", "))
-	}
-
 	var runtimeHelmCharts []string
 	relativeK8sPath := filepath.Join("var", "lib", "elemental", "kubernetes")
-	if needsHelmChartsSetup(&d.Kubernetes, m) {
+	if needsHelmChartsSetup(d, m) {
+		if len(d.Release.Enable) != 0 {
+			logger.Info("Enabling the following product extensions: %s", strings.Join(d.Release.Enable, ", "))
+		}
+
 		relativeHelmPath := filepath.Join(relativeK8sPath, "helm")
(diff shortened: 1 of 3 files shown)


examples/elemental/build/release.yaml
```diff
@@ -1,2 +1,4 @@
 name: suse-edge
 manifestURI: oci://registry.opensuse.org/home/adinov/core/containerfile/release-manifest:0.0.1
+enable:
+  - neuvector
```
atanasdinov, line 4: I can easily see this growing to have different types of enablements, for example:
```
enable:
   helm:
       - neuvector
    system:
       - dpdk
       - ...
```

However, it probably makes more sense to adapt to that closer to our "custom extensions" story rather than now.

ipetrov117 (commented):
Looks good 🚀 Left out a few comments and will be happy to approve once we discuss them.

examples/elemental/build/release.yaml
atanasdinov: In fact, this is very likely changing in an upcoming PR to allow for custom values.

ipetrov117 (approved):

davidcassany (approved):

examples/elemental/build/release.yaml
```diff
@@ -1,2 +1,4 @@
 name: suse-edge
 manifestURI: oci://registry.opensuse.org/home/adinov/core/containerfile/release-manifest:0.0.1
```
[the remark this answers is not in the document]
atanasdinov: It probably makes sense to have a "dummy" example in OBS (not IBS) just for the purpose of making the repository easy to use. Another idea is to have a local file in the same directory and have this field point to it via a `file://` reference.

PR #118, opened 2025-06-17
Refactor platform package

frelon:
Some of the functions still carry the Platform suffix which is already
made clear in the package name.

Also refactor the build-cmd to use platform instead of architecture to
be consistent.


internal/cli/elemental/cmd/build.go
```diff
@@ -47,10 +48,10 @@ func NewBuildCommand(appName string, action func(*cli.Context) error) *cli.Comma
...
 			},
 			&cli.StringFlag{
-				Name:        "arch",
-				Usage:       "Target architecture for the image (x86_64 or aarch64)",
-				Destination: &BuildArgs.Architecture,
-				Required:    true,
+				Name:        "platform",
+				Usage:       "Target platform",
+				Destination: &BuildArgs.Platform,
+				Value:       fmt.Sprintf("linux/%s", runtime.GOARCH),
```
fgiudici, line 54: nice

fgiudici (approved):
looks good here

internal/cli/elemental/cmd/build.go
atanasdinov, line 54: I'm very neutral about this, as it is going to be used by end users and also for grabbing the proper artefacts for airgapped builds where it is not unusual to have just the arch in their names.

The first concern might turn out not to be a problem, and the second one can be avoided with the proper helper functionality, so I'm good with this landing. We can always reevaluate later if necessary.

PR #119, opened 2025-06-18
Add coverprofile.out.bk to gitignore

frelon:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ go.work*
 
 # coverage report
 coverprofile.out
+coverprofile.out.bk
 
 # Go test files
 *.test


.gitignore
```diff
@@ -13,6 +13,7 @@ go.work*
 
 # coverage report
 coverprofile.out
+coverprofile.out.bk
```
ldevulder, line 16: Having `coverprofile.out*` take care of the 2 filenames (and more). Maybe better? Or maybe you want to explicitly exclude files?

frelon: Does not really matter to me, _coverprofile.out.bk_ is generated in the [Makefile](https://github.com/SUSE/elemental/blob/600ace5696a8aac96cc222e9762453daa921d542/Makefile#L55) so I added it explicitly..

davidcassany (approved):

PR #120, opened 2025-06-18
Add verify flag to install and upgrade command

frelon:
The verify flag is true by default, but can be set to false to enable
pulling images from registries that do not pass ssl verification.

### DIFF
--- a/pkg/transaction/transaction.go
+++ b/pkg/transaction/transaction.go
@@ -19,6 +19,7 @@ package transaction
 
 import (
 	"github.com/suse/elemental/v3/pkg/deployment"
+	"github.com/suse/elemental/v3/pkg/unpack"
 )
 
 type transactionState int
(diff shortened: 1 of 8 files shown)


pkg/transaction/transaction.go
```diff
@@ -53,7 +54,7 @@ type Interface interface {
 }
 
 type UpgradeHelper interface {
-	SyncImageContent(*deployment.ImageSource, *Transaction) error
+	SyncImageContent(*deployment.ImageSource, *Transaction, ...unpack.Opt) error
```
atanasdinov, line 57: Can these options be part of the struct itself in order not to modify the interface signature?

atanasdinov: Actually, it looks like it's a limitation of the `Init()` function of the other interface. Perhaps we could improve this in the future. 🤔

atanasdinov (approved):

pkg/transaction/transaction.go
frelon: Yep, this code is a bit convoluted and I will try to revisit this in the coming months!

PR #121, opened 2025-06-19
Add CreateBootEntry flag for upgrade

frelon:
The actual call to bootmanager.CreateBootEntries was never added in the
original firmware PR, add it now and add CreateBootEntry flag to Upgrade
command.


ipetrov117 (approved):

davidcassany (approved):
LGTM

PR #125, opened 2025-06-25
Replace deprecated gopkg.in/yaml.v3

frelon:
gopkg.in/yaml.v3 package has been unmaintained since april 2025 [1]

This commit replaces the usage of the package with the sigs.k8s.io/yaml
package used in other parts of the code-base.

[1] https://github.com/go-yaml/yaml/commit/944c86a7d29391925ed6ac33bee98a0516f1287a


go.mod
```diff
@@ -76,6 +75,7 @@ require (
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250218202821-56aae31c358a // indirect
 	google.golang.org/grpc v1.72.2 // indirect
 	google.golang.org/protobuf v1.36.6 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
fgiudici, line 78: Yeah, it will take time to fully get rid of it, since there are packages like jaypipes/ghw and gomega which import it.
Btw, great to at least start removing the direct calls 👍🏼

internal/image/install/install.go
```diff
@@ -18,6 +18,6 @@ limitations under the License.
 package install
 
 type Installation struct {
-	Bootloader    string `yaml:"bootloader"`
-	KernelCmdLine string `yaml:"kernelCmdLine"`
+	Bootloader    string `json:"bootloader"`
```
fgiudici, line 21: guess this is because sigs yaml package takes json annotations... so, the change of the package wasn't completely smooth, isn't it? 😆

fgiudici (approved):
Good catch!

frelon:
> @frelon tests on #127 seem to be passing, any idea why?

Interesting, that's also a way to do it! I would guess it's the difference between v2 and v3, my code changes uses `sigs.k8s.io/yaml` which wraps v2 and needs other changes for the tests to pass, but #127 uses v3 for the helm code which is pretty much the same code that was in `gopkg.in/yaml.v3`.

In the end I don't care which one we pick but we should use the same library in all of the code.

internal/image/install/install.go
frelon: Well, I saw that we already used the `json:` annotations elsewhere so I opted to use the same style in all of the code :sweat_smile:

go.mod
frelon: Yep, I was trying to `replace` this in the go-mod but couldn't find a way forward there..

pkg/deployment/deployment.go
```diff
@@ -163,73 +174,55 @@ func (p *PartRole) UnmarshalJSON(data []byte) (err error) {
...
-// as these are install|ugrade time information which is no longer guaranteed
-// to be (re)applicable on a deployed system as those are not kept aside.
-func (d Deployment) MarshalJSON() ([]byte, error) {
-	type deployAlias Deployment
-	deploy := deployAlias(d)
-	deploy.CfgScript = ""
-	deploy.OverlayTree = nil
-	return json.Marshal(deploy)
+	OverlayTree *ImageSource `yaml:"-"`
+	CfgScript   string       `yaml:"-"`
```
davidcassany, line 225: Isn't this a behavior change? Now this can only be set from the command line flags, not as part of a config yaml.

pkg/deployment/deployment.go
```diff
@@ -163,73 +174,55 @@ func (p *PartRole) UnmarshalJSON(data []byte) (err error) {
...
 
 type Partitions []*Partition
 
 type Disk struct {
-	Device      string     `json:"device,omitempty"`
-	Partitions  Partitions `json:"partitions"`
-	StartSector uint       `json:"startSector,omitempty"`
+	// omit the device name as this is a runtime information which might
+	// not be consistent across reboots, there is no need to store it.
+	Device      string     `yaml:"-"`
```
davidcassany, line 202: This forces setting the target device form the command line and not form config yaml.

davidcassany (commented):
I also noted with this change the target device can't be defined as part of the config file, neither the overlay and config script.

While it might not make a difference for our use cases I am a bit concerned about the lost flexibility. Specially regarding the target device, for which is not possible to set multiple disks at installation time with this change (imagine /var as a separate disk).

frelon:
> I also noted with this change the target device can't be defined as part of the config file, neither the overlay and config script.
> 
> While it might not make a difference for our use cases I am a bit concerned about the lost flexibility. Specially regarding the target device, for which is not possible to set multiple disks at installation time with this change (imagine /var as a separate disk).

I see, I did not know about this intended feature and it's not documented in any of the unit-tests. I will bring back the custom marshalling logic for this specific use-case.

pkg/deployment/deployment.go
frelon: Fixed in commit dc7da30a609950475c635da1048ad5fff397c142

davidcassany:
> I see, I did not know about this intended feature and it's not documented in any of the unit-tests. I will bring back the custom marshalling logic for this specific use-case.

Yes, fair point, all this is not really documented. I think having the option passing an installation setup as a yaml is great. It allows full flexibility for partitioning and its sizes, it might not be used in most cases, but I think it makes sense for more advanced use cases. I'd not close that door.

On the other side, the custom marshal logic is not intuitive, I agree on that. I'd be happy to consider other approaches, so far this is the simplest that came to my mind. The reasoning is simple, reading from yaml (e.g. reading a user provided file) unmarshals everything, but writing to a yaml (aka writing the /etc/elemental/deployment.yaml) omits runtime data.

PR #131, opened 2025-07-01
grub: Check if SLE theme directory exists

frelon:
Check that the ($root)/grub2/themes/SLE directory exists before trying
to load fonts and themes.

### DIFF
--- a/pkg/bootloader/grub.cfg
+++ b/pkg/bootloader/grub.cfg
@@ -95,13 +95,16 @@ done
 
 insmod part_gpt
 insmod gfxmenu
-loadfont ($root)/grub2/themes/SLE/DejaVuSans-Bold14.pf2
-loadfont ($root)/grub2/themes/SLE/DejaVuSans10.pf2
-loadfont ($root)/grub2/themes/SLE/DejaVuSans12.pf2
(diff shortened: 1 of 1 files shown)


pkg/bootloader/grub.cfg
```diff
@@ -95,13 +95,15 @@ done
...
-loadfont ($root)/grub2/themes/SLE/ascii.pf2
-insmod png
-set theme=($root)/grub2/themes/SLE/theme.txt
-export theme
+if [ -d "($root)/grub2/themes/SLE" ]; then
+  loadfont ($root)/grub2/themes/SLE/DejaVuSans-Bold14.pf2
+  loadfont ($root)/grub2/themes/SLE/DejaVuSans10.pf2
+  loadfont ($root)/grub2/themes/SLE/DejaVuSans12.pf2
+  loadfont ($root)/grub2/themes/SLE/ascii.pf2
+  insmod png
```
ldevulder, line 103: I think that `png` module can be loaded anyway, as we may have PNG  image even in TW distro.

frelon: Nice catch!

PR #132, opened 2025-07-02
Restructure build package

atanasdinov:
- Introduces a Builder type in order to simplify the build execution process
- Introduces a Helm type in order to abstract away the functionality around Helm charts
- Introduces an HTTP downloader logic that is going to be used across the various steps of the build process
- Splits up the Kubernetes configuration flow
- Adds wrappers around the build directory and known relative directory usages
- Adds unit tests for the large majority of the build functionality

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -49,12 +49,16 @@ var configScriptTpl string
 //go:embed templates/k8s_res_deploy.sh.tpl
 var k8sResDeployScriptTpl string
 
+type Builder struct {
+	System *sys.System
+	Helm   *Helm
+}
+
 // nolint:gocyclo
-func Run(ctx context.Context, d *image.Definition, buildDir string, valuesResolver helmValuesResolver, system *sys.System) error {
-	logger := system.Logger()
-	runner := system.Runner()
-	fs := system.FS()
-	overlaysPath := filepath.Join(buildDir, "overlays")
+func (b *Builder) Run(ctx context.Context, d *image.Definition, buildDir image.BuildDir) error {
+	logger := b.System.Logger()
+	runner := b.System.Runner()
+	fs := b.System.FS()
 
 	logger.Info("Resolving release manifest: %s", d.Release.ManifestURI)
 	m, err := resolveManifest(fs, d.Release.ManifestURI, buildDir)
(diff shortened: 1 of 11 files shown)


frelon:
> > Will it be possible to build a disk without actually including the hard-coded kubernetes/release-manifests?
> 
> I'd say that having a core release manifest defined will always be a requirement. Mainly because the `build` command uses the release-manifest to locate the OS and RKE2 images that it will use to build the actual image. While in the future it is very possible that we add a `--no-k8s`-like flag, I do not think that we would ever not want to include a release manifest.

In my opinion it should be the other way around, we should be able to build the smallest bootable artifact (without k8s) by default, and then k8s is just another extension on top. That way we don't have to rebuild everything when we have major changes, just modify which extensions we are using. If this is with/without a release manifest that is fine (we could even generate a minimal release-manifest if the user only wants to provide an OCI image).

> IMO, we have a two way approach:
> 
>     1. We have the `elemental install` command that allows for users to take an OCI image (that may be outside of UC's testing range) and do the necessary installation steps so that it is actually bootable - essentially what you have documented in [Add quick-start section to README #137](https://github.com/SUSE/elemental/pull/137)

Yes, but this is a very hacky approach and `install` should only be run from an installation media on an actual system to be installed, not a loopback. We probably want to add EFI variables on install by default which would break the host you are building on if you are not careful.

>     2. We have the `elemental build` command that allows for users to build a bootable, customised and extended image that is based on a specific manifest (e.g. UC core manifest, or Product extension manifest) that is validated
> 
> 
> To me these approaches are complimentary. Regarding (2), I'll try to add a generic enough documentation so that a user without access to IBS could follow along with the steps.

My point is that we are building a tool that is capable of both building a disk-image/installer from OCI, and then layering extensions on top. If we start to hard-code what extensions we are using we will always have new feature-requests to support new extensions, instead of taking a step back and architecting this to be used with any extension. 
Then the users can configure what they want without us having to change the code.

rdoxenham:
> > I'd say that having a core release manifest defined will always be a requirement. Mainly because the `build` command uses the release-manifest to locate the OS and RKE2 images that it will use to build the actual image. While in the future it is very possible that we add a `--no-k8s`-like flag, I do not think that we would ever not want to include a release manifest.
> 
> In my opinion it should be the other way around, we should be able to build the smallest bootable artifact (without k8s) by default, and then k8s is just another extension on top. That way we don't have to rebuild everything when we have major changes, just modify which extensions we are using. If this is with/without a release manifest that is fine (we could even generate a minimal release-manifest if the user only wants to provide an OCI image).
> 
> My point is that we are building a tool that is capable of both building a disk-image/installer from OCI, and then layering extensions on top. If we start to hard-code what extensions we are using we will always have new feature-requests to support new extensions, instead of taking a step back and architecting this to be used with any extension. Then the users can configure what they want without us having to change the code.

From my perspective, I totally agree. The default should be no Kubernetes, unless explicitly called. That way, the codebase can be used to generate Linux-only images, without being forced to explicitly disable Kubernetes. I can also imagine a situation where the `elemental3` API provides us with the ability to *not* specify a release manifest at all, and if they do that, they have to at least provide a Linux OCI as a starting point, and it can then optionally consume a `--config-dir`.

PR #135, opened 2025-07-03
Add --local flag to install, upgrade and build commands

davidcassany:
This flag forces the OCI unpack process to look for the given image in the local container storage. The library expects a daemon accessible through a socket typically defined by the DOCKER_HOST environment variable.

This commit also changes the behavior of attempting to find the image first in the local storage and fallback to remote if not present. With this change there is no fallback logic and pulling a remote or local image is always an explicit choice.

The PR is a follow up according to the discussion in https://github.com/SUSE/elemental/pull/106#pullrequestreview-2900406375

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -33,6 +33,7 @@ import (
 	"github.com/suse/elemental/v3/pkg/manifest/source"
 	"github.com/suse/elemental/v3/pkg/sys"
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
+	"github.com/suse/elemental/v3/pkg/unpack"
 	"github.com/suse/elemental/v3/pkg/upgrade"
 )
 
@@ -46,6 +47,7 @@ type Builder struct {
 	System       *sys.System
 	Helm         helmConfigurator
 	DownloadFile downloadFunc
+	Local        bool
 }
 
 func (b *Builder) Run(ctx context.Context, d *image.Definition, buildDir image.BuildDir) error {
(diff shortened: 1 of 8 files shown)


davidcassany:
This PR slightly overlaps with #132, but I don't think it should be a big problem to rebase one over the other, as this current one is small. So marking this as draft awaiting for #132 evolution.

Since there is #132 on the way I did not consider applying the `--local` flag to manifest extraction as this would imply passing values across several methods and structures in `elemental build` code path. So, for that particular corner case I believe it is better addressing #132 first.

atanasdinov (approved):
This can prove tricky to use down the road.

The main issue is the OS image itself is usually a big one, and we'd ideally not want to pull it on every build attempt. The unfortunate part is that this image is currently hidden in the release manifest i.e. in order to "cache" it, one would have to inspect a release manifest (and it gets even more complex if it's a product one), manually `podman pull` the OS image and proceed with `elemental build ... --local`.

We know that we are also looking to provide a ["Linux-only" build](https://github.com/SUSE/elemental/pull/132#issuecomment-3034868670) flow as well, so this might be fixed in the process or considered a valid preparation step. Still, further investigation might be required in order to simplify the UX once we get close to MVP.

ipetrov117 (approved):
I agree with @atanasdinov that further research is required on the caching topic and how it will relate to the `--local` flag. 

To me we should look to pull the image once and after that compare the digest of the pulled image with the digest of the remote one that is referred either from the manifest (for product extension use-cases) or from a cli flag specifying the desired OS image (for Linux-only build use-cases). For example, IIUC, we can use [remote.Get](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#Get) to get the image descriptor without pulling the actual image, meaning we could control when to pull the image and when to use the cached one based on the information of the descriptor.

Regarding for adding a local flag for the manifest extraction, I think that we should first focus on resolving the caching topic. Once this is done, it should be relatively simple to introduce local manifest image extraction.

Nevertheless, this is out of scope for this PR, to me this changes are a step in the right direction.

PR #136, opened 2025-07-03
Update openQA jobgroup config file

ldevulder:


test/cfg/openqa_elemental_jobgroup.yaml
```diff
@@ -31,18 +32,15 @@
   KEEP_GRUB_TIMEOUT: "0"
   TEST_PASSWORD: Elemental@R00t
   VIDEOMODE: "text"
+  REPO_TO_TEST: "https://dist.suse.de/ibs/Devel:/UnifiedCore:/Main:/ToTest/standard"
+  SYSEXT_PATH: "https://dist.suse.de/ibs/Devel:/UnifiedCore:/Main:/ToTest/sysexts/"
```
dirkmueller, line 36: is it intentional that one of the two ends with a `/` and the other one does not?

ldevulder: For now yes it is, otherwise the `curl` command inside the code will not be able to get the sysexts' list. But I think I will force a `/` inside the test code to avoid this but in a next change.

frelon (approved):

test/cfg/openqa_elemental_jobgroup.yaml
ldevulder: In fact I made the fix directly here: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/22551. Cleaner and better to do it now.

PR #139, opened 2025-07-10
Add basic integration test

ldevulder:
This PR adds all the basics needed to be able to schedule integration tests running on a VM with a Tumbleweed based OS.

Most of the scripts/helpers are backported from `rancher/elemental-toolkit`. I tried to keep only what we need to avoid unneeded code.

As this PR is already big enough there is only a basic integration test, it will be improved in another PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ GO_BUILD_ARGS?=-ldflags '$(LDFLAGS)'
 DOCKER?=docker
 DISKSIZE?=20G
 OS_REPO?=registry.opensuse.org/devel/unifiedcore/tumbleweed/containers/uc-base-os-kernel-default
-OS_VERSION?=0.0.1
+OS_VERSION?=latest
 ELEMENTAL_IMAGE_REPO?=local/elemental-image
 DOCKER_SOCK?=/var/run/docker.sock
 ifdef PLATFORM


pkg/transaction/snapper.go
```diff
@@ -203,7 +203,10 @@ func (sn *snapperT) probe(d deployment.Deployment) (err error) {
 
 	part := sn.hwPartitions.GetByUUIDNameOrLabel(sysPart.UUID, sysPart.Role.String(), sysPart.Label)
 	if part == nil {
-		return fmt.Errorf("system partition not found: %+v", sysPart)
+		part, err = block.GetPartitionByUUID(sn.s, blockDev, sysPart.UUID, 4)
```
ldevulder, line 206: The first version made by @frelon completely replaced `GetByUUIDNameOrLabel` call by `GetPartitionByUUID`, it works but I had failures in the Unit tests and I didn't find how to fix them...

This method makes the tests happy and also I think it's better to use use `GetPartitionByUUID` only when needed.

pkg/bootloader/grub.cfg
```diff
@@ -43,19 +35,18 @@ if [ "${prev_saved_entry}" ]; then
 fi
 
 function savedefault {
-  if [ -z "${boot_once}" ]; then
+  if ! test -n "${boot_once}"; then
```
ldevulder, line 38: According to Grub documentation `-z` does not exist.

tests/vm/sut.go
```diff
@@ -0,0 +1,280 @@
...
+	"fmt"
+	"log"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2" //nolint:staticcheck,revive
+	. "github.com/onsi/gomega"    //nolint:staticcheck,revive
```
atanasdinov, line 30: Having to do this is not ideal. Is there a better approach?

tests/vm/sut.go
```diff
@@ -0,0 +1,280 @@
...
+		actual := s.BootFrom()
+		if actual != image {
+			return fmt.Errorf("expected boot from %s, actual %s", image, actual)
+		}
+
+		return nil
+	}, time.Duration(60)*time.Second, time.Duration(10)*time.Second).ShouldNot(HaveOccurred())
+}
+
+func (s *SUT) GetOSRelease(ss string) string {
```
atanasdinov, line 143: I'm conflicted. We perform assertions within the scope of these functions and in turn our tests look like they're not doing anything:
```
		It("Extract informations from booted OS", func() {
			osVersion := s.GetOSRelease("VERSION_ID")
			osArch := s.GetArch()

			By(fmt.Sprintf("OS Version: %s", osVersion))
			By(fmt.Sprintf("OS Architecture: %s", osArch))
		})
```

Can we refactor this so that SUT functions return values or errors and then adapt the tests to include these assertions?

atanasdinov (commented):
Thanks for kicking this off! It is going to prove really valuable with time.

I do have some suggestions and questions. Nothing is really critical but I'm concerned few places are not going to produce valuable error messages (or logs) and I'm afraid I don't really understand the functionality SUT implements while not being part of a `_test.go` file.

Makefile
```diff
@@ -46,6 +78,28 @@ $(BUILD_DIR)/elemental3: $(GO_FILES)
 $(BUILD_DIR)/elemental3-toolkit: $(GO_FILES)
 	go build $(GO_BUILD_ARGS) -o $@ ./cmd/elemental-toolkit
 
+.PHONY: toolkit-container
+toolkit-container:
+	$(DOCKER) build --platform $(PLATFORM) --target runner --tag $(TOOLKIT_REPO):$(VERSION) .
+
+.PHONY: build-disk
+build-disk: $(BUILD_DIR) toolkit-container
```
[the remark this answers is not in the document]
ldevulder: What do you mean? The `toolkit-container` target (will be renamed as you suggested) includes the elemental binaries and use them to create a disk image. I choose this solution instead of executing directly on the GitHub runner as it's Ubuntu OS and all our tests are done mainly on (open)SUSE distribution.

pkg/block/device.go
```diff
@@ -139,6 +145,22 @@ func GetPartitionByUUID(s *sys.System, b Device, uuid string, attempts int) (*Pa
 	return nil, errors.New("no device found")
 }
 
+func lookupUUID(s *sys.System, path string) string {
+	out, err := s.Runner().Run("blkid", path)
+	if err != nil {
+		s.Logger().Error("blkid execution failed: %s", out)
```
ldevulder, line 151: I think that `Error` here is a bit too much. It doesn't break anything but seeing ` time="2025-07-21T14:32:04Z" level=error msg="failed parsing blkid output"` is a bit annoying. `Warning` should be enough I think, @davidcassany WDYT?

tests/vm/sut.go
```diff
@@ -0,0 +1,294 @@
...
+		pass = "uc0@linux"
+	}
+
+	if host = os.Getenv("SSH_HOST"); host == "" {
+		host = "127.0.0.1:2222"
+	}
+
+	vmPidStr := os.Getenv("VM_PID")
+	if value, err = strconv.Atoi(vmPidStr); err == nil {
+		By(fmt.Sprintf("Underlaying VM pid is set to: %d", value))
```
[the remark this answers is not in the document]
davidcassany: This was mostly included to debug the VM runner script, it took me a while get it right so having some pid traces helped me coding it. I guess we can remove it now.

davidcassany (commented):
Great job. I think this is a great base setting the environment to easily and gradually expand.

As mentioned on some of my comments I have concerns on the changes related to hardware detection... these are behavior changes and it is not clear to me why these are required. To me it feels like a workaround more than a fix.

examples/elemental/install/config.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -xe
+
+# Disable Grub timeout
+grub2-editenv /boot/grubenv set timeout=5
+grub2-editenv /boot/grubenv set extra_cmdline=console=ttyS0,115200
```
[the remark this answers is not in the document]
ldevulder: Oh yes it could be! TBH I was not aware of that `--cmdline` option... Either I missed it or it wasn't not available at the beginning. But anyway I will replace mu specific code with this one. Easier and it does exactly the same!

pkg/bootloader/grub.cfg
```diff
@@ -122,20 +113,20 @@ else
...
-        linux "${linux}" ${cmdline}
-        echo 'Loading initial ramdisk ...'
-        initrd "${initrd}"
-    }
+  load_env --file (${root})/loader/entries/${entry}
+
+  menuentry "${display_name}" --id "${entry}" {
+    load_env --file (${root})/loader/entries/${chosen}
+    echo 'Loading Linux...'
+    linux "${linux}" ${cmdline} ${extra_cmdline}
```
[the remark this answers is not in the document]
ldevulder: Yes will use this!

tests/vm/sut.go
ldevulder: I fixed this a bit by checking the returned values. And For now it's a very simple test that needs to be extended and fixed. So I think that I will remove this specific block later.

tests/vm/ssh.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	client := ssh.NewClient(c, chans, reqs)
+
+	go func() {
+		t := time.NewTicker(2 * time.Second)
+		defer t.Stop()
+		for range t.C {
+			_, _, err := client.SendRequest("keepalive@golang.org", true, nil)
+			if err != nil {
+				return
```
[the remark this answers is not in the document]
ldevulder: I tested and this add lot of mess on the test logs for nothing and it's just a keep-alive "error", so not very useful IMHO.

Makefile
```diff
@@ -1,25 +1,47 @@
...
 
 LDFLAGS:=-w -s
 LDFLAGS+=-X "$(GO_MODULE)/internal/cli/version.version=$(GIT_TAG)"
 LDFLAGS+=-X "$(GO_MODULE)/internal/cli/version.gitCommit=$(GIT_COMMIT)"
 
 GO_BUILD_ARGS?=-ldflags '$(LDFLAGS)'
 
+# Used to build the OS image
+DOCKER?=docker
+FLAVOR?=green
```
davidcassany, line 31: nit: I am not sure we need the FLAVOR concept here... Probably this can be dropped or set to `tumbleweed`. I don't think we need a distinction such as green, orange or blue in here.

pkg/transaction/snapper.go
ldevulder: As we discussed: removed!

pkg/transaction/snapper.go
```diff
@@ -241,19 +244,21 @@ func (sn *snapperT) probe(d deployment.Deployment) (err error) {
...
 		return fmt.Errorf("creating partition mountpoint path '%s': %w", mountPoint, err)
 	}
-	bPart := sn.hwPartitions.GetByUUID(part.UUID)
-	if bPart == nil {
-		return fmt.Errorf("partition '%s' not found", part.UUID)
-	}
-	err = sn.s.Mounter().Mount(bPart.Path, mountPoint, "", []string{"rw"})
+
+	bDev := lsblk.NewLsDevice(sn.s)
+	bPart, err := block.GetPartitionByUUID(sn.s, bDev, part.UUID, 4)
```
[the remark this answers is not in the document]
ldevulder: As we discussed: removed!

Makefile
```diff
@@ -46,6 +78,28 @@ $(BUILD_DIR)/elemental3: $(GO_FILES)
...
+
+.PHONY: build-disk
+build-disk: $(BUILD_DIR) elemental-image
+	qemu-img create -f raw $(IMG).raw $(DISKSIZE)
+	TARGET=$$(sudo losetup -f --show $(IMG).raw)
+	cp examples/elemental/install/config.sh $(BUILD_DIR)
+	$(DOCKER) run --rm \
+		--volume $(DOCKER_SOCK):$(DOCKER_SOCK) \
+		--volume $(BUILD_DIR):/build \
+		--volume /dev:/dev \
```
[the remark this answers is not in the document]
ldevulder: Done and it works, thanks!

davidcassany (approved):
Looks great! I am impressed by the runtime, just 4m! :+1:

PR #141, opened 2025-07-11
Configure network

atanasdinov:


internal/build/network.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/suse/elemental/v3/internal/image"
+	"github.com/suse/elemental/v3/internal/template"
+	"github.com/suse/elemental/v3/pkg/sys/vfs"
+)
+
+//go:embed templates/network.sh.tpl
+var configureNetworkScript string
+
+const networkCustomScriptName = "configure-network.sh"
```
ipetrov117, line 33: nit:  Feel free to ignore this if you do not agree, but would it make sense to rename this to `configNetworkScriptName` to avoid the confusion that this constant is somehow connected to the one defined in the [parseNetwork](https://github.com/atanasdinov/elemental/blob/configure-network/internal/cli/elemental/action/build.go#L211) function of the build action? IIUC this property enforces the name of the script (be it custom or default) as seen in the OS, while the other enforces the name of the custom script that the user can provide in the config directory. Which while somewhat connected, describe and enforce different things.

internal/build/templates/network.sh.tpl
```diff
@@ -0,0 +1,20 @@
...
+set -euo pipefail
+
+MARKER_FILE="/var/lib/elemental/.network-configuration-attempted"
+
+if [ -f "$MARKER_FILE" ]; then
+    echo "Marker file '$MARKER_FILE' found. Script already executed, exiting." >&2
+    exit 0
+fi
+
+/usr/bin/nmc apply --config-dir {{ .ConfigDir }} || {
```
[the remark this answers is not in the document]
ipetrov117: Sounds good, wanted to make sure, as I am not familiar with the internals of `nmc` 👍

ipetrov117 (approved):

internal/build/templates/network.sh.tpl
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
+
+MARKER_FILE="/var/lib/elemental/.network-configuration-attempted"
```
dirkmueller, line 4: should this be elemental3?

internal/build/templates/config.sh.tpl
```diff
@@ -8,6 +8,32 @@ useradd -m {{ .Username }} || true
...
+{{- if .NetworkScript }}
+# Configuring network
+
+cat <<- EOF > /etc/systemd/system/first-boot-network.service
+[Unit]
+Description=Configure network on first boot
+Before=first-boot-complete.target
+Wants=first-boot-complete.target
+
+Before=rke2-server.service rke2-agent.service
```
dirkmueller, line 20: If I understand https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html correctly, before means 

>After= ensures the opposite, that the listed unit is fully started up before the configured unit is started

as it is `type=oneshot` that means immediately after the network.sh has finished running. however, I think at that point nm hasn't finished reconfiguring the network, has it? which means rke2 is potentially running already while network is being reconfigured?

atanasdinov: >  If unit foo.service contains the setting Before=bar.service and both units are being started, bar.service's start-up is delayed until foo.service has finished starting up. 

While clearly not sufficient, the network configuration script finishes execution very fast and I wasn't able to observe RKE2 starting during its execution, hence I never ran into issues. Obviously, this is not a guarantee it will always work but it will have to be addressed at a later point as discussed earlier.

rdoxenham: We absolutely need to make sure that we can guarantee that network setup runs before RKE2 starts. The moment that certs are generated and etcd members are established on an incorrect setup (e.g. IP addresses received by DHCP), we cannot easily change them.

PR #150, opened 2025-07-22
Introduce network configuration docs

ipetrov117:
Introduces documentation for the changes that where introduced as part of #141.


.github/workflows/unit_tests.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
           go-version-file: go.mod
       - name: Install dependencies
         run: |
+          sudo apt-get update
```
ipetrov117, line 27: Mitigates [404 Not Found](https://github.com/SUSE/elemental/actions/runs/16448737324/job/46534945717) errors when installing the `rootlesskit` package.

PR #168, opened 2025-09-01
Implicitly enable Kubernetes by specifying workloads

atanasdinov:
This PR adjusts the current modification flow, such that:
- requirements to specify the origin of the enabled release components is now dropped
  - "product" extensions always have a priority over the "core" ones
- Kubernetes is now only enabled if at least one of the following is configured
  - Kubernetes resources and manifests under `/kubernetes/manifests`
  - Kubernetes manifests or Helm charts under `kubernetes.yaml`
  - Helm chart extensions under `release.yaml`

In the future, especially once validation for the input is added, we might add an additional (explicit) approach to going for the Linux-only builds.

Additionally, the documentation is only briefly edited to adhere to the new changes. As soon as OS extensions are further fleshed out and integrated into the build flow, these are going to undergo the respective editing.


davidcassany (commented):
Looks good :+1:

internal/build/helm.go
```diff
@@ -185,21 +159,48 @@ func (h *Helm) collectHelmCharts(charts []helmChart, repositories, valueFiles ma
...
+		}
+	}
+
+	if rm.ProductExtension != nil && rm.ProductExtension.Components.Helm != nil {
+		for _, c := range rm.ProductExtension.Components.Helm.Charts {
+			productCharts[c.Chart] = c
+		}
+
+		for _, repository := range rm.ProductExtension.Components.Helm.Repositories {
+			repositories[repository.Name] = repository.URL
```
ipetrov117, line 185: Do we want to also safeguard agains consumers defining a repository name that has already been defined from the `corePlatform`? IIUC, as it is right now, consumers can override repository URLs by just specifying the same name, but a different value. Granted this can be considered a feature, but we need to be aware of it.

ipetrov117 (commented):
Looks good to me, one minor question and I'll be happy to approve.

internal/build/helm.go
ipetrov117: Understood. We could definitely approach the problem by setting up, for example, prefixes to the `corePlatform` repositories (e.g. `core-<repo_name>`) and instructing consumers/customers to not name their repositories in that manner.

However, I do think we should validate this. Perhaps not here exactly, but have some validation that checks if a specific consumer/customer repository name starts with an unsupported prefix. That said, this does not need to happen in this PR for sure.

ipetrov117 (approved):

internal/build/kubernetes.go
```diff
@@ -33,8 +33,16 @@ import (
...
+func needsManifestsSetup(def *image.Definition) bool {
+	return len(def.Kubernetes.RemoteManifests) > 0 || len(def.Kubernetes.LocalManifests) > 0
+}
+
+func needsHelmChartsSetup(def *image.Definition) bool {
+	return (len(def.Release.Components.Helm) > 0) || def.Kubernetes.Helm != nil
+}
+
+func isKubernetesEnabled(def *image.Definition) bool {
+	return needsHelmChartsSetup(def) || needsManifestsSetup(def)
```
frelon, line 45: Reading this I guess it's not possible to deploy an empty rke2 cluster?

Should we treat rke2 as it's own component to be able to deploy a box with "just" linux+kubernetes  and no workloads?

atanasdinov: Currently not possible, no.

The issue with that is there are already too many juggling parts and it feels too early to impose restrictions, e.g. if we have an entry for it in `release.yaml` or it is a separate flag, then the omission of this but inclusion of a Kubernetes manifest or a Helm chart will either have to fail (which is an inconvenient UX) or implicitly pass (which basically means the entry / flag is not respected at all).

atanasdinov: This was discussed offline and we decided to proceed with the current changes. Future iterations will improve on extensions and "empty" Kubernetes enablement.

PR #175, opened 2025-09-10
Add --enable-fips flag for install command

frelon:
The --enable-fips flag will set the boot= and fips=1 kernel parameters
as well as running `fips-mode-setup --enable --no-bootcfg` during the
install procedure.

The running system shows the following output:

```
FIPS mode is enabled.
Initramfs fips module is enabled.
The current crypto policy (FIPS) is based on the FIPS policy.
```

### DIFF
--- a/pkg/bootloader/grub.go
+++ b/pkg/bootloader/grub.go
@@ -329,12 +329,14 @@ func (g *Grub) installKernelInitrd(rootPath, espDir, subfolder, snapshotID, kern
 	}
 
 	// Copy kernel .hmac in order to enable FIPS.
-	kernelHmac := filepath.Join(filepath.Dir(kernel), ".vmlinuz.hmac")
-	if exists, _ := vfs.Exists(g.s.FS(), kernelHmac); exists {
-		err = vfs.CopyFile(g.s.FS(), kernelHmac, targetDir)
-		if err != nil {
-			return nil, fmt.Errorf("copying kernel hmac '%s': %w", kernelHmac, err)
-		}
+	kernelHmac, err := vfs.FindKernelHmac(g.s.FS(), kernel)
+	if err != nil {
+		return nil, fmt.Errorf("finding kernel hmac '%s': %w", kernel, err)
+	}
+
+	err = vfs.CopyFile(g.s.FS(), kernelHmac, targetDir)
+	if err != nil {
+		return nil, fmt.Errorf("copying kernel hmac '%s': %w", kernelHmac, err)
 	}
 
 	initrdPath := filepath.Join(filepath.Dir(kernel), Initrd)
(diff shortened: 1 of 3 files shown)


pkg/bootloader/grub.go
```diff
@@ -328,6 +328,15 @@ func (g *Grub) installKernelInitrd(rootPath, espDir, subfolder, snapshotID, kern
 		return nil, fmt.Errorf("copying kernel '%s': %w", kernel, err)
 	}
 
+	// Copy kernel .hmac in order to enable FIPS.
+	kernelHmac := filepath.Join(filepath.Dir(kernel), ".vmlinuz.hmac")
```
dirkmueller, line 332: this should also check for `.Image.hmac` (for aarch64).

frelon: Updated to use the kernel filename to look for the .hmac file.

frelon:
Linting fixed (I added `//nolint:gocyclo` to the upgrade method, since refactoring that method seems out of scope for this PR.)

Also updated the .hmac copying in bootloader to use the previously found kernel filename.

pkg/deployment/deployment.go
```diff
@@ -211,11 +211,16 @@ type FirmwareConfig struct {
...
+type FipsConfig struct {
+	Enabled bool `yaml:"enabled"`
+}
+
 type Deployment struct {
 	SourceOS   *ImageSource    `yaml:"sourceOS"`
 	Disks      []*Disk         `yaml:"disks"`
 	Firmware   *FirmwareConfig `yaml:"firmware"`
 	BootConfig *BootConfig     `yaml:"bootloader"`
+	Fips       *FipsConfig     `yaml:"fips"`
```
atanasdinov, line 223: Do we expect more fields? If not, is a simple pointer to boolean enough?

internal/cli/action/install.go
```diff
@@ -136,7 +136,16 @@ func digestInstallSetup(s *sys.System, flags *cmd.InstallFlags) (*deployment.Dep
...
-		d.BootConfig.KernelCmdline = flags.KernelCmdline
+		d.BootConfig.KernelCmdline = fmt.Sprintf("%s %s", d.BootConfig.KernelCmdline, flags.KernelCmdline)
+	}
+
+	if flags.EnableFips {
+		d.Fips = &deployment.FipsConfig{
+			Enabled: true,
+		}
+
+		bootFlag := fmt.Sprintf("boot=LABEL=%s", deployment.EfiLabel)
```
atanasdinov, line 147: Is this affected by FIPS enablement? I assumed we always need it.

pkg/fips/fips.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	"github.com/suse/elemental/v3/pkg/chroot"
+	"github.com/suse/elemental/v3/pkg/sys"
+)
+
+func Enable(ctx context.Context, s *sys.System) error {
+	stdOut, err := s.Runner().RunContext(ctx, "/usr/bin/fips-mode-setup", "--enable", "--no-bootcfg")
```
atanasdinov, line 28: I believe this is decommissioned [upstream](https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/merge_requests/205#9ac97f30b729f8d1c92231a3f266fc15fbf8c8c6).

We are a couple of commits behind on [SLFO:Main](https://build.suse.de/projects/SUSE:SLFO:Main/packages/crypto-policies/files/crypto-policies.spec?expand=1) hence why the script is still present.

dirkmueller: It also changes cryptopolicies. calling the command is still the recommended way for SL16, even if upstream decided to change things.

PR #184, opened 2025-09-18
Add --install-* parameters to customize command

frelon:
This commit adds the same --install-* flags from the build-iso command
to the customize command.

It reuses the same code and maps the overlay, config-script and
generated install.yaml into /Install in the customized ISO.

This enables to customize the installed kernel cmdline as well as other
installed artifacts (config-script, overlay etc).


davidcassany (approved):
Nice :+1: This makes way easier to create autoinstaller ISOs from a generic ISO.

PR #187, opened 2025-09-26
Support systemd extensions from OCI images

atanasdinov:
This allows the integration of systemd extensions delivered in OCI format in release manifests now, and as separate user additions later.

### DIFF
--- a/internal/build/systemd_sysext.go
+++ b/internal/build/systemd_sysext.go
@@ -20,6 +20,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"path/filepath"
 	"slices"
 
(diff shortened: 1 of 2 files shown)


internal/build/systemd_sysext.go
```diff
@@ -46,19 +48,36 @@ func (b *Builder) downloadSystemExtensions(ctx context.Context, def *image.Defin
...
+			if err = b.DownloadFile(ctx, fs, extension.Image, extensionPath); err != nil {
+				return fmt.Errorf("downloading systemd extension %s: %w", extension.Name, err)
+			}
 
-		if err = b.DownloadFile(ctx, fs, extension.Image, extensionPath); err != nil {
-			return fmt.Errorf("downloading systemd extension %s: %w", extension.Name, err)
+			continue
+		}
+
+		unpacker := unpack.NewOCIUnpacker(b.System, extension.Image, unpack.WithLocalOCI(b.Local))
```
atanasdinov, line 63: I contemplated using a generic unpacker here but that would increase the wrapping by one level, e.g. if someone wants to provide an extension from a local directory, they'd have to create their extension as a directory or a disk first, and then embed that in another directory here. The confusion for disks would be even bigger in the sense that regardless of their origin (remote or local), we'd expect them to be ready to use as is and not attempt to unpack them.

Considering that this feature will firstly be used only for release manifests, I figured we can start out with OCI, and then revisit once we start thinking on the user (not consumer!) extensions story.

davidcassany (approved):
Looks good to me. I wonder if we might need some fine tuning depending on how we finally pack sysext in OCI images (fs image embeeded, just the directory tree, sboms and related metadata, etc.), but I we guess we will just discover it once we get there.

PR #200, opened 2025-10-15
Switch to systemd-repart to handle disk partitioning

davidcassany:
This commit drops the parted and sgdisk partitioner implementations and moves to systemd-repart. Main reasons are dropping a relevant chunk of code, simplify the repart logic and facilitate introducing additional features already supported by systemd-repart (luks, dm-verity, etc.)

There are a couple of behavior changes:

* UUIDs of partitions now refer to the GPT UUID, not the filesystem UUID
* There is no longer control of the start sector of each partition

### DIFF
--- a/pkg/block/lsblk/lsblk.go
+++ b/pkg/block/lsblk/lsblk.go
@@ -167,7 +167,7 @@ func (l lsDevice) GetDeviceSectorSize(device string) (uint, error) {
 		return 0, fmt.Errorf("no sector size reported by lsblk %v", device)
 	}
 
-	return unmarshalSectorSize(out)
+	return size, err
 }
 
 // GetPartitionFS gets the filesystem type for the given partition device. If the given device
--- a/pkg/sys/runner/runner.go
+++ b/pkg/sys/runner/runner.go
@@ -65,7 +65,7 @@ func (r run) RunEnv(command string, env []string, args ...string) ([]byte, error
 	out, err := cmd.Output()
 	if err != nil {
 		r.debug("'%s' command reported an error: %s", command, err.Error())
-		r.debug("'%s' command output: %s", command)
+		r.debug("'%s' command output: %s", command, out)
 	}
 	return out, err
 }


davidcassany:
Some further insights and reasoning around this PR.

Why?

Started looking at systemd-repart in order to create raw disks without having to mount any FS (aka unprivileged build) and realized I could easily use it for the installation too. Does this allow unprivileged installations? NO, in this PR systemd-repart is only used to create GPT partitions and format them, it doesn't feed them with any data, hence no extra memory is required during install and we still need to mount the FSs to actually sync data in them.

Behavior changes:

* GPT labels and Filesystem labels are always the same value. Before this PR only filesystem labels were configurable and GPT labels were simply hardcoded and known values. This changes, now both are configurable but always assigned with the same value, no distinction. In classic Elemental we used to use GPT labels as a reliable way to identify partitions and allow FS labes customization. In elemental 3 I do not see this as a need and for a similar purpose we have now discoverable partitions spec.
* UUIDs in this PR refer to the partition UUIDs, not the filesystem UUID. This PR makes partition UUIDs configurable but not the FS UUID. Which is assigned randomly. Moreover system-repart support discoverable partitions meaning that undefined partition UUID are not randomly assigned, but set to the known values to determine the role and purpose of the partition, see [discoverable partitions](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/).
* With this PR we can't control the start sector anymore. Why this was introduced at first place (this was not even part of classic Elemental)? Because some ARM boards have hilarious requirements on that regard, such as ESP partition needs to start at nth sector or some magical blob needs to be copied at certain sector. I know, at the very least, this is required by some rockchip based boards. In that regard, two considerations would be, first, probably not a wise choice being concerned about these corner cases now, and, second, I'd assume standardization in ARM world is just happening and this might not be an issue on modern and industrial hardware. In any case two workarounds are identified, we could set GPT table before calling systemd-repart with a custom first usable sector, and second workaround, we could create an empty 1st partition to add extra offset. So IMHO this first sector configuration loss is not an issue.

rdoxenham:
Thanks for the great work on this, @davidcassany -- you're right to call out the start sector, some embedded boards do call the bootloader at a very specific sector of the disk. If we lose the ability to do this without a workaround, can we document that workaround, or provide an option for users to specify this via the schema?

pkg/block/lsblk/lsblk.go
```diff
@@ -118,14 +143,33 @@ func (l lsDevice) GetAllPartitions() (block.PartitionList, error) {
...
+	size, err := unmarshalSectorSize(out)
+	if err != nil {
+		return 0, err
+	}
+
+	if size == 0 {
+		return 0, fmt.Errorf("no sector size reported by lsblk %v", device)
+	}
+
+	return unmarshalSectorSize(out)
```
atanasdinov: This is called twice. Perhaps leftover from debugging?

davidcassany:
rebased from main

ipetrov117 (approved):

PR #203, opened 2025-10-23
Refactor installer package

davidcassany:
This PR includes several changes to facilitate disk builds in follow up PRs. To build an installer disk it is expected to build a pretty similar system as the installer ISO, in that regard, the expectation is to generalize further the code of former `installermedia` package.

Overview of changes included in the PR:
* `installermedia` package renamed to `installer` only to keep it shorter. Also avoid referring to ISO in favor of Live media or Installer.
* some of the fields of the ISO struct are moved to the Deployment object. This is to make the Deployment object the main input to describe the actual outcome, regardless if we are building a expanding disk, an ISO (including the description of the system to install) or running a full clean installation.
* Make install process check if running from an ISO and, if so, then try to fetch the deployment description embedded in the ISO. This allows to install from an installer media by just running `elemental3ctl install --target <device>` and the installation will honor any customization applied on the ISO.
* Improve sanity checks in Deployment to catch missing OS source image or missing disk device.

### DIFF
--- a/internal/cli/action/build_iso.go
+++ b/internal/cli/action/build_iso.go
@@ -99,7 +99,7 @@ func digestInstallerDeploymentSetup(s *sys.System, label string, flags *cmd.Inst
 
 	err = applyInstallFlags(s, d, flags.InstallSpec)
 	if err != nil {
-		return nil, fmt.Errorf("failed applying install flags to deployment description")
+		return nil, fmt.Errorf("failed applying install flags to deployment description: %w", err)
 	}
 
 	err = d.Sanitize(s, deployment.CheckDiskDevice)
--- a/internal/cli/action/customize.go
+++ b/internal/cli/action/customize.go
@@ -83,7 +83,7 @@ func digestCustomizeDeploymentSetup(s *sys.System, flags *cmd.CustomizeFlags) (*
 
 	err := applyInstallFlags(s, d, flags.InstallSpec)
 	if err != nil {
-		return nil, fmt.Errorf("failed applying install flags to deployment description")
+		return nil, fmt.Errorf("failed applying install flags to deployment description: %w", err)
 	}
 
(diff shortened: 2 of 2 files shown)


davidcassany:
> I think that this PR has an impact on the current docs: do you planned to update it in this PR or in another (as this one is already big)?

This has no relevant impact in docs, however we can simplify a bit the installer ISO example in docs. Gonna add it in this PR. Thanks for raising the concern.

internal/cli/action/customize.go
```diff
@@ -66,19 +63,37 @@ func Customize(ctx *cli.Context) error {
...
+		d.Installer.CfgScript = flags.ConfigScript
+	}
+
+	if flags.KernelCmdline != "" {
+		d.Installer.KernelCmdline = flags.KernelCmdline
+	}
+
+	err := applyInstallFlags(s, d, flags.InstallSpec)
+	if err != nil {
+		return nil, fmt.Errorf("failed applying install flags to deployment description")
```
atanasdinov:
```suggestion
		return nil, fmt.Errorf("failed applying install flags to deployment description: %w", err)
```

internal/cli/action/build_iso.go
```diff
@@ -78,25 +72,44 @@ func BuildInstaller(ctx *cli.Context) error { //nolint:dupl
...
+
+	src, err := deployment.NewSrcFromURI(flags.OperatingSystemImage)
+	if err != nil {
+		return nil, fmt.Errorf("invalid OS image URI (%s) to build installer: %w", flags.OperatingSystemImage, err)
 	}
+	d.SourceOS = src
 
+	err = applyInstallFlags(s, d, flags.InstallSpec)
+	if err != nil {
+		return nil, fmt.Errorf("failed applying install flags to deployment description")
```
atanasdinov:
```suggestion
		return nil, fmt.Errorf("failed applying install flags to deployment description: %w", err)
```

frelon (approved):

PR #215, opened 2025-10-30
Replace qemu-img with truncate in build

frelon:
This removes a dependency on the qemu-img package during building. Truncate is installed from coreutils and available on most systems.

Noticed this dependency in #214


ldevulder (approved):
LGTM.

The only advantages to have `qemu-img` would have been to be able to directly use/generate `qcow2` image. But I think it's more generic and better to use `raw` and do conversion after yes.

So I will remove this deps from my doc PR as well ;-)

atanasdinov (approved):

PR #219, opened 2025-11-06
Dynamically reload kernel modules

atanasdinov:
Introduces a new **kmod** CLI command to elemental3ctl that allows:
* Reloading additional kernel modules that are coming from systemd extensions (sysexts)
* Unloading said modules in order to establish proper cleanup procedures once sysexts are unmerged

The initial proposal was to bind mount the module cache directory directly against the kernel directory (`/usr/lib/modules/<kernel-version>`) in order to "mask" the various `modules.*` files. Whilst technically working, this included a circular dependency as the effective "lower dir" for both the cache overlay and the bind mount were the exact same path, i.e. the kernel directory itself.

Instead, this implementation retains the original plan for module cache overlay but it dynamically executes `modprobe(8)` against the result. In order to achieve this we define a stable mount point at `/run/elemental/kmod`.

The logic for creating an overlayFS itself is extracted to its own package as it could be useful in the future. It is easily extendable in case a specific use case requires additional options.

**This implementation relies on the fact that every Elemental 3 host (or Kubernetes node) ships the elemental3ctl systemd extension.**

The CLI command is currently plugged-in as part of the `ensure-sysext.service` oneshot. As soon as `systemd-sysext.service` runs and merges the contents of all extensions on the filesystem, elemental3ctl runs the reload operation to inject any merged modules to the kernel. The aforementioned mount point remains intact even after the command finishes. Once `systemd-sysext.service` is restarted (e.g. via `systemctl restart`), this implementation will pick up on the "Stop" event and trigger the unloading operation so that any modules injected from the current extension set is removed from the kernel and once the extensions are reloaded, the reload operation will kick in to reestablish whatever is necessary.

Snippet from the logs on what this looks like:
```
node1:~ # systemctl restart systemd-sysext
node1:~ # journalctl -u ensure-sysext
...
Nov 06 01:36:35 node1.qemu systemd[1]: Stopping Reload systemd to include units provided by sysexts...
Nov 06 01:36:35 node1.qemu sh[12495]: time="2025-11-06T01:36:35Z" level=info msg="Module aesni_intel deactivating completed"
Nov 06 01:36:35 node1.qemu sh[12495]: time="2025-11-06T01:36:35Z" level=info msg="Module kvm_intel deactivating completed"
Nov 06 01:36:35 node1.qemu sh[12495]: time="2025-11-06T01:36:35Z" level=info msg="Unloading kernel modules completed"
Nov 06 01:36:35 node1.qemu systemd[1]: ensure-sysext.service: Deactivated successfully.
Nov 06 01:36:35 node1.qemu systemd[1]: Stopped Reload systemd to include units provided by sysexts.
Nov 06 01:36:36 node1.qemu systemd[1]: Starting Reload systemd to include units provided by sysexts...
Nov 06 01:36:36 node1.qemu sh[12644]: time="2025-11-06T01:36:36Z" level=info msg="Generating module files"
Nov 06 01:36:37 node1.qemu sh[12644]: time="2025-11-06T01:36:37Z" level=info msg="Activating kernel modules"
Nov 06 01:36:37 node1.qemu sh[12644]: time="2025-11-06T01:36:37Z" level=info msg="Module aesni_intel activating completed"
Nov 06 01:36:37 node1.qemu sh[12644]: time="2025-11-06T01:36:37Z" level=info msg="Module kvm_intel activating completed"
Nov 06 01:36:37 node1.qemu sh[12644]: time="2025-11-06T01:36:37Z" level=info msg="Reloading kernel modules completed"
Nov 06 01:36:37 node1.qemu systemd[1]: Finished Reload systemd to include units provided by sysexts.
```

All kernel modules that are part of this flow have to be defined as part of their respective extensions in the core or product release manifests. They are then converted to `/etc/elemental/deployment.yaml` entries during the customisation process. The ones in the logs above are found on Elemental 3 systems and were manually appended to the `/etc/elemental/deployment.yaml` file for simplicity. Additional tests were also performed with an NVIDIA extension and, aside from a GPU missing report, everything appears to be working as intended.

Finally, it's important to note that there is a lot of potential improvements for this feature -- this is just the first draft! On top of my head we could:
- Allow the ability to specify additional modules @ runtime that are not part of `/etc/elemental/deployment.yaml`
- Design a more generic API where the feature is not so keen on the `/run/elemental/kmod` mount point
- Introduce additional capabilities such as querying the kernel modules from extensions and their current status
- Continuously retry modprobe calls or establish priority handling in case module dependencies cause transient errors
- Gracefully handle the various potential depmod and modprobe failures

### DIFF
--- a/internal/build/templates/ensure-sysext.service
+++ b/internal/build/templates/ensure-sysext.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=Reload systemd units and kernel modules provided by sysexts
+Description=Reload systemd units provided by sysexts
 BindsTo=systemd-sysext.service
 After=systemd-sysext.service
 DefaultDependencies=no
@@ -9,14 +9,11 @@ ConditionDirectoryNotEmpty=|/run/extensions
 ConditionDirectoryNotEmpty=|/var/lib/extensions
 ConditionDirectoryNotEmpty=|/usr/local/lib/extensions
 ConditionDirectoryNotEmpty=|/usr/lib/extensions
-ConditionPathExists=/usr/bin/elemental3ctl
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/systemctl daemon-reload
-# TODO (atanasdinov): Investigate a SELinux limitation where elemental3ctl cannot be called directly here
-ExecStart=/bin/sh -c "/usr/bin/elemental3ctl kmod --reload"
 ExecStart=/usr/bin/systemctl restart --no-block sockets.target timers.target multi-user.target
 
 [Install]
--- a/internal/build/ignition.go
+++ b/internal/build/ignition.go
@@ -36,13 +36,17 @@ import (
 )
 
 const (
-	ensureSysextUnitName = "ensure-sysext.service"
-	k8sResourcesUnitName = "k8s-resource-installer.service"
-	k8sConfigUnitName    = "k8s-config-installer.service"
+	ensureSysextUnitName        = "ensure-sysext.service"
+	reloadKernelModulesUnitName = "reload-kernel-modules.service"
+	k8sResourcesUnitName        = "k8s-resource-installer.service"
+	k8sConfigUnitName           = "k8s-config-installer.service"
 )
 
 //go:embed templates/ensure-sysext.service
 var ensureSysextUnit string
+
+//go:embed templates/reload-kernel-modules.service
+var reloadKernelModulesUnit string
 
 //go:embed templates/k8s-resource-installer.service.tpl
 var k8sResourceUnitTpl string
@@ -90,7 +94,6 @@ func (b *Builder) configureIgnition(def *image.Definition, buildDir image.BuildD
 			return err
 		}
 
-		config.AddSystemdUnit(ensureSysextUnitName, ensureSysextUnit, true)
 		config.AddSystemdUnit(k8sResourcesUnitName, k8sResourcesUnit, true)
 	}
 
@@ -111,6 +114,9 @@ func (b *Builder) configureIgnition(def *image.Definition, buildDir image.BuildD
 			Path:     extensions.File,
 			Contents: v0_6.Resource{Inline: util.StrToPtr(data)},
 		})
+
+		config.AddSystemdUnit(ensureSysextUnitName, ensureSysextUnit, true)
+		config.AddSystemdUnit(reloadKernelModulesUnitName, reloadKernelModulesUnit, true)
 	}
 
 	ignitionFile := filepath.Join(buildDir.FirstbootConfigDir(), image.IgnitionFilePath())
--- a/internal/build/ignition_test.go
+++ b/internal/build/ignition_test.go
@@ -141,6 +141,8 @@ passwd:
 		Expect(err).NotTo(HaveOccurred())
 
 		Expect(ignition).To(ContainSubstring("/etc/elemental/extensions.yaml"))
+		Expect(ignition).To(ContainSubstring("Reload systemd units"))
+		Expect(ignition).To(ContainSubstring("Reload kernel modules"))
 		Expect(ignition).NotTo(ContainSubstring("merge"))
 		Expect(ignition).NotTo(ContainSubstring("Kubernetes Resources Installer"))
 		Expect(ignition).NotTo(ContainSubstring("Kubernetes Config Installer"))
--- a/internal/build/templates/reload-kernel-modules.service
+++ b/internal/build/templates/reload-kernel-modules.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Reload kernel modules provided by sysexts
+After=systemd-sysext.service
+DefaultDependencies=no
+ConditionPathExists=/usr/bin/elemental3ctl
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+# TODO (atanasdinov): Investigate a SELinux limitation where elemental3ctl cannot be called directly here
+ExecStart=/bin/sh -c "/usr/bin/elemental3ctl kmod --reload"
+TimeoutStartSec=1min
+
+[Install]
+WantedBy=sysinit.target


internal/build/templates/ensure-sysext.service
```diff
@@ -9,12 +9,16 @@ ConditionDirectoryNotEmpty=|/run/extensions
...
+ConditionPathExists=/usr/bin/elemental3ctl
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/systemctl daemon-reload
+# TODO (atanasdinov): Investigate a SELinux limitation where elemental3ctl cannot be called directly here
+ExecStart=/bin/sh -c "/usr/bin/elemental3ctl kmod --reload"
 ExecStart=/usr/bin/systemctl restart --no-block sockets.target timers.target multi-user.target
+ExecStop=/bin/sh -c "/usr/bin/elemental3ctl kmod --unload"
```
rdoxenham: I'm torn a little on this; do we explicitly want to unload modules without rebooting into the *new* ones (or *older* ones if it's a rollback)? I'm trying to figure out the ideal UX here, whether it makes sense to only load, and assume that any reload operations are part of a reboot, where we also refresh the OS image. What use-case(s) do I need to be thinking about here?

internal/kmod/kmod.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, fmt.Errorf("deployment information not found")
+	}
+
+	var kernelModules []string
+
+	if d.Extensions != nil {
+		kernelModules = d.Extensions.KernelModules
+	}
+
+	// TODO (atanasdinov): Merge modules from deployment.yaml with ones coming from CLI args?
```
rdoxenham: In what sense? Are you thinking that the *originally* supplied kernel modules from `deployment.yaml` need to be appended to the list of modules now being supplied via sysext? Why wouldn't these match?

internal/kmod/unload.go
```diff
@@ -0,0 +1,54 @@
...
+	logger := u.Logger()
+
+	isMountPoint, _ := u.Mounter().IsMountPoint(u.Config.MountPoint)
+	if !isMountPoint {
+		logger.Warn("Kernel modules mount point not found, skipping unload")
+		return nil
+	}
+
+	if err := manageModules(ctx, u.System, u.Config.MountPoint, kernelModules, true); err != nil {
+		logger.Error("Deactivating one or more kernel modules failed")
```
rdoxenham: I think this will certainly need to be handled well... there will absolutely be scenarios where you cannot unload a kernel module cleanly, for example, it's in-use, or another module is depending on it. We cannot expect to be able to unload when we want to, hence my earlier comment about always assuming that the unloading and reloading of kernel modules would be via a reboot, i.e. because we've upgraded (or rolled back) a system.

internal/build/templates/ensure-sysext.service
```diff
@@ -9,12 +9,16 @@ ConditionDirectoryNotEmpty=|/run/extensions
...
 ConditionDirectoryNotEmpty=|/usr/local/lib/extensions
 ConditionDirectoryNotEmpty=|/usr/lib/extensions
+ConditionPathExists=/usr/bin/elemental3ctl
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/bin/systemctl daemon-reload
+# TODO (atanasdinov): Investigate a SELinux limitation where elemental3ctl cannot be called directly here
+ExecStart=/bin/sh -c "/usr/bin/elemental3ctl kmod --reload"
```
dirkmueller: can we move this to its own systemd service rather than overloading `ensure-sysext` with it? as I fixed the reload manager handling (with vojtechs help), this entire service should actually be removeable..

atanasdinov: We can do the cleanup in a follow-up, though!

davidcassany:
> Help me understand the suggestion for providing the users tooling, please. Is that an expectation on their side to run a script, is it a systemd unit that we embed or something else entirely?

This can be many things. It could be a simple unit & configfile (to list kernel modules) included as part of the ignition configuration. Elemental build could craft it as it already does with other configuration options and services. From `elemental build` user interface perspective nothing changes.

Lets try to visit my concern from another perspective.

What currently has elemental3ctl as input:

* disk layout (input at upgrade|install|build-installer)
* OS image (input at upgrade|install|build-installer)
* bootloader/firmware setup (input at install|upgrade|build-installer time)
* config script (executed at install|upgrade time or included at build-installer time)
* overlay (drop-in files install|upgrade|build-installer time)

Note elemental3ctl as of today has nothing to do with extensions. They are not even assumed anywhere as of today.

With this PR a list of kernel modules list to modporbe at boot time id added to the above list, with the implicit assumption they are provided by sysext extensions. I struggle to see it fits there. Because it configures boot time operations and because if feels a very specific case data, data I'd encompass in a wider extensions management story.

I can image evolving elemental3ctl to be aware of sysext and to be capable to install|upgrade them. Which would relate well with the idea of aligning as much as possible different scenarios and use cases. Then we could explore means to define extensions, its metadata and special cases if any (elemental3ctl, kernel modules, services, etc.).

So what if we keep `elemental build` responsible of setting the boot time operations and then potentially expand elemental3ctl capabilities to handle sysexts on a later iteration when the full lifecycle management comes into play. What you think?

> We could end up with elemental3ctl regenerating modules at a specific directory, then expect a third component, large script or plain documentation so that users rely on said directory, is arguably a more fragile implementation.

I see this point. I imagined it a bit different, as entirely separate of elemental3ctl extension. An approach that has other down sides too like having to handle additional pieces somewhere else or having to manage huge boot time setup in `elemental build`. In any case I am not concerned about making modules cache regeneration as part of `elemental3ctl` or not, if this is only about regenerating the modules cache to prefixed location this is a pretty isolated task.

>The proposed draft minimises moving parts, acknowledges a decoupling of exact extension contents by relying on the preconfigured deployment during customisation, and ensures seamless automation for the user. Please let me know what you think on the above -- I'd be more than happy to discuss further and meet somewhere in the middle.

Sure, please consider my comments as suggestions, I am not opposing to the current PR.

internal/build/templates/ensure-sysext.service
```diff
@@ -9,12 +9,15 @@ ConditionDirectoryNotEmpty=|/run/extensions
 ConditionDirectoryNotEmpty=|/var/lib/extensions
 ConditionDirectoryNotEmpty=|/usr/local/lib/extensions
 ConditionDirectoryNotEmpty=|/usr/lib/extensions
+ConditionPathExists=/usr/bin/elemental3ctl
```
davidcassany: This is implicitly making elemental3ctl extension a hard requirement to load any service provided by an extension. I agree with @dirkmueller that `elemental3ctl kmod` should be handled separately in its own unit and do not have any kind dependency with `ensure-sysext` and only depend on `systemd-sysext` service, in fact, the two could run in parallel.

I am fine leaving it like this for now, but IMHO this requires  follow up PR.

PR #230, opened 2025-11-13
Install RKE2 SELinux module on firstboot

atanasdinov:
RKE2 is unable to start properly if it is enabled with the `selinux: true` configuration option. This is due to the systemd extension, containing both RKE2 itself and rke2-selinux policy, coming with properly relabeled files, resulting in improper context for the containerd and containerd-shim processes. By manually injecting the selinux policy this is now fixed.

### DIFF
--- a/internal/build/templates/k8s-config-installer.service.tpl
+++ b/internal/build/templates/k8s-config-installer.service.tpl
@@ -8,6 +8,8 @@ Type=oneshot
 TimeoutSec=900
 Restart=on-failure
 RestartSec=60
+# TODO (atanasdinov): Figure out a declarative, non-hardcoded approach for installing selinux modules
+ExecStartPre=/bin/sh -c "semodule -i /usr/share/selinux/packages/rke2.pp"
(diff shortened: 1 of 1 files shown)


internal/build/templates/k8s-config-installer.service.tpl
```diff
@@ -8,6 +8,8 @@ Type=oneshot
 TimeoutSec=900
 Restart=on-failure
 RestartSec=60
+# TODO (atanasdinov): Fix context of files and directories instead of only forcing a rebuild and reload of policies
+ExecStartPre=/bin/sh -c "semodule -B"
```
dirkmueller: would `semodule --refresh` be sufficient? e.g. why do we need an unconditional rebuild and reload?

atanasdinov:
This is not enough to fix the issue, as even rebuilding the policy still results in failures for sysexts that have removed the `Upholds` clause for the `multi-user.target`. Currently investigating, thanks @ipetrov117 for finding this.

PR #235, opened 2025-11-14
Adding installer disk builder

davidcassany:
This PR refactors installer package to be able to build ISOs and disks. Disks are defined with a very similar setup to ISOs,  they only include the EFI and the recovery partitions.

Also `build-iso` command is renamed to `build-installer` as this can be used for both, ISOs and disks. This rename also makes prominent the purpose of the built image.


davidcassany:
This PR is missing docs for the new command, I am on it. Code wise is reviewable

atanasdinov (approved):

ldevulder (approved):

PR #237, opened 2025-11-17
Reset command to expand disk

davidcassany:
This PR adds the reset command which installs the system from the recovery partition to the rest of the current disk. As of today it can only reset if the rest of the disk is not partitioned yet.

Running from a installer disk image calling `elemental reset` is enough to self install and expand to the current disk geometry. No further flags are required as all the installation details were already embedded as part of the build installer process.


davidcassany:
I am well aware this is missing docs update, gonna update them as  followup PR.

davidcassany:
Rebased, added some docs example and reduced duplication between install and reset actions

PR #241, opened 2025-11-18
Remove build-dir after build

frelon:
The `build` command leaves artifacts in the build-dir that the find
command later complains about when running `make`:

```
find: ‘./build/build-2025-11-18T12-35-00’: Permission denied
find: ‘./build/build-2025-11-18T12-35-00’: Permission denied
```

This commit adds a defer method in the build action that removes the
build-dir after the build is done.

### DIFF
--- a/internal/cli/action/build.go
+++ b/internal/cli/action/build.go
@@ -76,6 +76,14 @@ func Build(ctx *cli.Context) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


ldevulder (approved):

atanasdinov:
Let's reevaluate once there are useful log files to store.

PR #242, opened 2025-11-18
Install HA metallb setup for kubernetes-svc

frelon:
In order to achieve high-availability for the rke2 server nodes we deploy metallb and expose a LoadBalancer for the kubernetes-service.

This enables us to deploy multiple control-plane servers and load-balance the kubernetes API using metallb l2 mode [1]

A view of the deployed resource:

```
$ kubectl get svc --all-namespaces
NAMESPACE        NAME                                      TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                         AGE
default          kubernetes                                ClusterIP      10.43.0.1       <none>          443/TCP                         5m5s
default          kubernetes-vip                            LoadBalancer   10.43.119.206   10.0.2.15       9345:32173/TCP,6443:30453/TCP   2m28s
```

In order to sync the kubernetes service in case a node goes down we also deploy the endpoint-copier which will copy endpoints from the kubernetes service to kubernetes-vip. This works around an issue with using metallb for load-balancing the kubernetes API [2].

[1] https://metallb.universe.tf/concepts/layer2/

### DIFF
--- a/internal/build/ignition.go
+++ b/internal/build/ignition.go
@@ -42,17 +42,22 @@ const (
 	k8sConfigUnitName           = "k8s-config-installer.service"
 )
 
-//go:embed templates/ensure-sysext.service
-var ensureSysextUnit string
+var (
+	//go:embed templates/ensure-sysext.service
+	ensureSysextUnit string
 
-//go:embed templates/reload-kernel-modules.service
-var reloadKernelModulesUnit string
+	//go:embed templates/reload-kernel-modules.service
+	reloadKernelModulesUnit string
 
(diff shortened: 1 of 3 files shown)


frelon:
> While this includes the resources that we need on the cluster, there is no logic to automatically deploy MetalLB and Endpoint Copier Operator here. If the user does not provide those, then the multi-node setup will fail.
> 
> Can we look into fetching those from the release manifest based on the criteria of `apiVIP` being provided?

Good point, I added some code to automatically deploy those charts in case they are not set in the `release.yaml` :+1:

atanasdinov (approved):

PR #246, opened 2025-11-19
Add --log-file global CLI flag

frelon:
The --log-file flag takes a string that can be a path to a file or
-/stdout/stderr and sets the log target.

### DIFF
--- a/internal/cli/cmd/root.go
+++ b/internal/cli/cmd/root.go
@@ -18,19 +18,31 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/urfave/cli/v2"
 
 	"github.com/suse/elemental/v3/pkg/log"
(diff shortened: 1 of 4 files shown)


internal/cli/cmd/root.go
```diff
@@ -44,9 +56,42 @@ func Setup(ctx *cli.Context) error {
...
+
 	if ctx.App.Metadata == nil {
 		ctx.App.Metadata = map[string]any{}
 	}
 	ctx.App.Metadata["system"] = s
 	return nil
 }
+
+func Teardown(ctx *cli.Context) error {
+	return logFile.Close()
```
atanasdinov: This is going to panic with nil dereference in the cases where we don't log to file. Perhaps a single `if logFile != nil` first should be enough?

atanasdinov (approved):

PR #248, opened 2025-11-19
Enhancements to the deployment merge function

ipetrov117:
This PR attempts to fix the below described problems, while also providing additional unit test. 

_Problem:_
1. When the source deployment (e.g. a deployment coming from `elmenetal3 customize --install-description`) defines a partition slice that is smaller than the partition slice of the dest deployment (e.g. an `install.yaml` file from an ISO installer media) the dest deployment partition slice is disregarded and only the src partition slice is present.
2. As stated in the [code](https://github.com/SUSE/elemental/blob/main/pkg/deployment/merge.go#L42-L47) Disks and Partitions are merged one by one in order. That is fine for Disks, however when merging partitions we get the following problems:
   1. Entries that are present in the destination deployment, but are not present in the source deployment will be replaced.
   2. Merging partitions one by one without checking what we are merging causes different partitions to be merged together which breaks the deployment state.

Example for problem (1):

1. `install.yaml` file in an ISO installer before customization:
   ```yaml
   ...
   disks:
     - partitions:
       - label: EFI
          fileSystem: vfat
          size: 1024
          role: efi
          mountPoint: /boot
          mountOpts:
            - defaults
            - x-systemd.automount
        - label: RECOVERY
          fileSystem: btrfs
          size: 1280
          role: recovery
          hidden: true
        - label: SYSTEM
          fileSystem: btrfs
          role: system
          mountPoint: /
          mountOpts:
            - ro=vfs
          rwVolumes:
            - path: /var
              noCopyOnWrite: true
              mountOpts:
                - x-initrd.mount
            - path: /root
              mountOpts:
                - x-initrd.mount
            - path: /etc
              snapshotted: true
              mountOpts:
                - x-initrd.mount
            - path: /opt
            - path: /srv
            - path: /home
    ...
   ```

1. Source deployment that is passed to the installer media using `elemental3 customize --install-description deployment.yaml ..`:
   ```yaml
   disks:
    - target: /dev/sda
      partitions: []
   ```
1. After booting the customized installer media, the merged `install.yaml` deployment file looks like this:
   ```yaml
   ...
   disks: 
     - target: /dev/sda
       partitions: []
   ...
   ```
1. Which in turn produces the following error in the `elemental-autoinstall.service`:
   ```
   Nov 19 15:57:32 localhost.localdomain elemental3ctl[1452]: 2025/11/19 15:57:32 inconsistent deployment setup found: no 'system' partition defined
   ```

Example for problem (2):

1. `install.yaml` file in an ISO installer before customization is the same as in problem (1) (Note that the `RECOVERY` partition is defined).
1. Source deployment that is passed to the installer media using `elemental3 customize --install-description deployment.yaml ..`:
   ```yaml
   disks:
    - target: /dev/sda
      partitions:
        - label: EFI
          fileSystem: vfat
          size: 1024
          role: efi
          mountPoint: /boot
          mountOpts:
            - defaults
            - x-systemd.automount
        - label: ignition
          fileSystem: btrfs
          size: 256
          role: data
          mountPoint: /run/elemental/firstboot
          hidden: true
        - label: ELEMENTAL-PREPARE
          fileSystem: btrfs
          size: 128
          role: data
          mountPoint: /run/elemental/prepare
          hidden: true
        - label: SYSTEM
          fileSystem: btrfs
          role: system
          mountPoint: /
          mountOpts:
            - ro=vfs
          rwVolumes:
            - path: /var
              noCopyOnWrite: true
              mountOpts:
                - x-initrd.mount
            - path: /root
              mountOpts:
                - x-initrd.mount
            - path: /etc
              snapshotted: true
              mountOpts:
                - x-initrd.mount
            - path: /opt
            - path: /srv
            - path: /home
   ```
1. After booting the customized installer media, the merged `install.yaml` deployment file looks like this:
   ```yaml
   ...
   disks:
     - target: /dev/sda
        partitions:
          - label: EFI
            fileSystem: vfat
            size: 1024
            role: efi
            mountPoint: /boot
            mountOpts:
              - defaults
              - x-systemd.automount
          - label: ignition
            fileSystem: btrfs
            size: 256
            role: data
            mountPoint: /run/elemental/firstboot
            hidden: true
          - label: ELEMENTAL-PREPARE
            fileSystem: btrfs
            size: 128
            role: data
            mountPoint: /run/elemental/prepare
            mountOpts:
              - ro=vfs
            rwVolumes:
              - path: /var
                noCopyOnWrite: true
                mountOpts:
                  - x-initrd.mount
              - path: /root
                mountOpts:
                  - x-initrd.mount
              - path: /etc
                snapshotted: true
                mountOpts:
                  - x-initrd.mount
              - path: /opt
              - path: /srv
              - path: /home
            hidden: true
          - label: SYSTEM
            fileSystem: btrfs
            role: system
            mountPoint: /
            mountOpts:
              - ro=vfs
            rwVolumes:
              - path: /var
                noCopyOnWrite: true
                mountOpts:
                  - x-initrd.mount
              - path: /root
                mountOpts:
                  - x-initrd.mount
              - path: /etc
                snapshotted: true
                mountOpts:
                  - x-initrd.mount
              - path: /opt
              - path: /srv
              - path: /home
   ...
   ```
1. Two are the things of note here:
   1. The `RECOVERY` partition that was present in the destination deployment is now missing - This happens because the `ignition` partition essentially overrides the `RECOVERY` partition, as partitions with index 1 are merged.
   2. The `ELEMENTAL-PREPARE` partition from the source was merged with the `SYSTEM` partition from dest deployment - This happens due to the merge of partitions with index 2 from both the src (`ELEMNTAL-PREPARE`) and dest (`SYSTEM`) deployments.
1. Finally this results in the following error on the `elemental-autoinstall.service`:
   ```
   Nov 19 16:07:36 localhost.localdomain elemental3ctl[1415]: 2025/11/19 16:07:36 inconsistent deployment setup found: rw volume paths must be unique. Duplicated '/var'
   ```

_Proposed Solution:_
1. Partitions with the same label in both source and destination deployments are merged
2. Non-merged partitions (e.g. labeled partitions that are present in either source or destination) will be appended to the final result
3. Partitions without a label will be appended to the final result.
4. Partitions with duplicate labels from source will have only the last duplicate partition merged to destination (e.g. if 3 partitions have the same label in source, only the last defined of those 3 partitions will be merged with a partition with the same label from destination).

Example with proposed solution:

1. `install.yaml` file in an ISO installer before customization is the same as in problem (1)
2. `deployment.yaml` source file is the same as in problem (2)
3. After installer media customization the merged `install.yaml` file looks like this:
```yaml
...
  disks:
    - target: /dev/sda
      partitions:
        - label: ignition
          fileSystem: btrfs
          size: 256
          role: data
          mountPoint: /run/elemental/firstboot
          hidden: true
        - label: ELEMENTAL-PREPARE
          fileSystem: btrfs
          size: 128
          role: data
          mountPoint: /run/elemental/prepare
          hidden: true
        - label: EFI
          fileSystem: vfat
          size: 1024
          role: efi
          mountPoint: /boot
          mountOpts:
            - defaults
            - x-systemd.automount
        - label: RECOVERY
          fileSystem: btrfs
          size: 1280
          role: recovery
          hidden: true
        - label: SYSTEM
          fileSystem: btrfs
          role: system
          mountPoint: /
          mountOpts:
            - ro=vfs
          rwVolumes:
            - path: /var
              noCopyOnWrite: true
              mountOpts:
                - x-initrd.mount
            - path: /root
              mountOpts:
                - x-initrd.mount
            - path: /etc
              snapshotted: true
              mountOpts:
                - x-initrd.mount
            - path: /opt
            - path: /srv
            - path: /home
...
```

This essentially ensures that all partitions are merged accordingly to their counterparts, or just appended if said counterparts do not exit.

### DIFF
--- a/pkg/deployment/deployment_test.go
+++ b/pkg/deployment/deployment_test.go
@@ -25,7 +25,6 @@ import (
 	. "github.com/onsi/gomega"
 	"go.yaml.in/yaml/v3"
 
-	"github.com/suse/elemental/v3/pkg/bootloader"
 	"github.com/suse/elemental/v3/pkg/deployment"
 	"github.com/suse/elemental/v3/pkg/log"
 	"github.com/suse/elemental/v3/pkg/sys"
@@ -195,49 +194,6 @@ var _ = Describe("Deployment", Label("deployment"), func() {
 			_, err := deployment.Parse(s, "/some/dir")
 			Expect(err).NotTo(HaveOccurred())
 			Expect(buffer.String()).To(ContainSubstring("deployment file not found"))
-		})
-		It("merges the default deployment with a given one", func() {
-			d := deployment.New(
-				deployment.WithPartitions(1, &deployment.Partition{
-					Role:      deployment.Recovery,
-					Label:     deployment.RecoveryLabel,
-					Size:      2048,
-					MountOpts: []string{"defaults", "ro"},
-				}),
-				deployment.WithConfigPartition(1024),
-			)
-			d.SourceOS = deployment.NewEmptySrc()
-			newD := &deployment.Deployment{
-				SourceOS: deployment.NewOCISrc("domain.org/image/repo:tag"),
-				Disks: []*deployment.Disk{
-					{Device: "/dev/device", Partitions: []*deployment.Partition{
-						{Size: 2048}, nil, {Label: "newLabel", Size: 4096},
-						{MountPoint: "/data", MountOpts: []string{"rw"}},
-					}}, nil,
-				},
-				CfgScript: "script",
-				BootConfig: &deployment.BootConfig{
-					KernelCmdline: "new cmdline",
-				},
-			}
-			Expect(deployment.Merge(d, newD)).To(Succeed())
-
-			Expect(d.SourceOS.String()).To(Equal("oci://domain.org/image/repo:tag"))
-			Expect(d.CfgScript).To(Equal("script"))
-			Expect(len(d.Disks)).To(Equal(1))
-			Expect(d.Disks[0].Device).To(Equal("/dev/device"))
-			Expect(len(d.Disks[0].Partitions)).To(Equal(3))
-			Expect(d.Disks[0].Partitions[0].Size).To(Equal(deployment.MiB(2048)))
-			system := d.GetSystemPartition()
-			Expect(system).NotTo(BeNil())
-			Expect(system.Size).To(Equal(deployment.MiB(0)))
-			Expect(d.Disks[0].Partitions[1].Role).To(Equal(deployment.Recovery))
-			Expect(d.Disks[0].Partitions[1].Label).To(Equal("newLabel"))
-			Expect(d.Disks[0].Partitions[1].Size).To(Equal(deployment.MiB(4096)))
-			Expect(system.MountOpts).To(Equal([]string{"rw"}))
-			Expect(len(system.RWVolumes)).To(Equal(6))
-			Expect(d.BootConfig.KernelCmdline).To(Equal("new cmdline"))
-			Expect(d.BootConfig.Bootloader).To(Equal(bootloader.BootNone))
 		})
 		It("deep copies the current deployment", func() {
 			d := deployment.DefaultDeployment()
--- a/pkg/deployment/merge_test.go
+++ b/pkg/deployment/merge_test.go
@@ -1,0 +1,377 @@
+/*
+Copyright © 2025 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deployment_test
+
+import (
+	"fmt"
+	"strconv"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/suse/elemental/v3/pkg/deployment"
+)
+
+var _ = Describe("Deployment merge", Label("deployment"), func() {
+	It("merges an empty source", func() {
+		src := &deployment.Deployment{}
+		dst := &deployment.Deployment{
+			Disks: []*deployment.Disk{
+				{
+					Device: "/dev/sda",
+				},
+			},
+		}
+
+		Expect(deployment.Merge(dst, src)).To(Succeed())
+		Expect(len(dst.Disks)).To(Equal(1))
+		Expect(dst.Disks[0].Device).To(Equal("/dev/sda"))
+	})
+
+	It("merges an empty src partition", func() {
+		src := &deployment.Deployment{
+			Disks: []*deployment.Disk{
+				{
+					Partitions: deployment.Partitions{},
+				},
+			},
+		}
+		dst := &deployment.Deployment{
+			Disks: []*deployment.Disk{
+				{
+					Partitions: deployment.Partitions{
+						{
+							Label: "SYSTEM",
+						},
+					},
+				},
+			},
+		}
+
+		Expect(deployment.Merge(dst, src)).To(Succeed())
+		Expect(len(dst.Disks)).To(Equal(1))
+		Expect(len(dst.Disks[0].Partitions)).To(Equal(1))
+		Expect(dst.Disks[0].Partitions[0].Label).To(Equal("SYSTEM"))
+	})
+
+	It("merges deployment disks", func() {
+		src := &deployment.Deployment{
+			Disks: []*deployment.Disk{
+				{
+					Device: "/dev/sda",
+				},
+			},
+		}
+		dst := &deployment.Deployment{
+			Disks: []*deployment.Disk{
+				{
+					Partitions: deployment.Partitions{
+						{
+							Label: "SYSTEM",
+						},
+					},
+				},
+			},
+		}
+
+		Expect(deployment.Merge(dst, src)).To(Succeed())
+		Expect(len(dst.Disks)).To(Equal(1))
+		Expect(len(dst.Disks[0].Partitions)).To(Equal(1))
+		Expect(dst.Disks[0].Partitions[0].Label).To(Equal("SYSTEM"))
+		Expect(dst.Disks[0].Device).To(Equal("/dev/sda"))
+	})
+
+	It("merges deployment partitions", func() {
+		src := &deployment.Deployment{
+			Disks: []*deployment.Disk{
+				{
+					Partitions: deployment.Partitions{
+						{
+							Label:     "SYSTEM",
+							MountOpts: []string{"ro=vfs"},
+						},
+					},
+				},
+			},
+		}
+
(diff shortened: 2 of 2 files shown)


pkg/deployment/merge.go
```diff
@@ -39,26 +39,41 @@ func (t *transformer) Transformer(typ reflect.Type) func(dest, src reflect.Value
...
-// Merge applies non zero values of the source deployment to the destination deployment.
-// The list of disks and partitions are merged by order. For instance, first source partition
-// is merged with first destination partition. If source disks or partitions list is longer
-// than the destination one, any overflow item is appended.
+// Merge applies non zero values of src to dst.
 //
-// Other slice types in Deployment are simply replaced, not merged.
+// Supported merge slice types include: Disks and Partitions.
+// Disks are merged by order (e.g. first instance of src is merged with first instance of dst).
+// Partitions are merged by label (e.g. label "foo" from src will only be merged with label "foo" from dst).
```
davidcassany: Why by label and not UUID, Roles, partition label or any other of the many attributes a partition has? What if there are duplicated or crossed UUIDs?

I am not a fan of identifying partitions by filesystem label, it tends to be not robust enough.

davidcassany (commented):
> 1. When the source deployment (e.g. a deployment coming from `elmenetal3 customize --install-description`) defines a partition slice that is smaller than the partition slice of the dest deployment (e.g. an `install.yaml` file from an ISO installer media) the dest deployment partition slice is disregarded and only the src partition slice is present.

This is what is fixing #244

> 2. As stated in the [code](https://github.com/SUSE/elemental/blob/main/pkg/deployment/merge.go#L42-L47) Disks and Partitions are merged one by one in order. That is fine for Disks, however when merging partitions we get the following problems:
> 
>     1. Entries that are present in the destination deployment, but are not present in the source deployment will be replaced.
> 
>     2. Merging partitions one by one without checking what we are merging causes different partitions to be merged together which breaks the deployment state.

I don't think this is an issue, but a feature. The caller should be aware of the layout it wants and the source of truth is the partitions order. If the caller wants to relabel system partition with ROOT label then this is possible, just keep empty (not nil) any of the previous partitions and merge will keep them and then at the system's partition position just add a partition only with the label defined. If you need to add a new partition somewhere in the middle then you need to provide the full layout, but still possible.

> Two are the things of note here:
>
>    The RECOVERY partition that was present in the destination deployment is now missing - This happens because the ignition partition essentially overrides the RECOVERY partition, as partitions with index 1 are merged.
>    The ELEMENTAL-PREPARE partition from the source was merged with the SYSTEM partition from dest deployment - This happens due to the merge of partitions with index 2 from both the src (ELEMNTAL-PREPARE) and dest (SYSTEM) deployments.

Yes, this is the intended behavior, order matters. There is no partition matching logic to keep merge logic simple, otherwise trying to be smart and matching partitions has high potential to hit corner cases because of the matching criteria (trying to relabel an already defined partition, changing UUIDS, partitions without a filesystem label defined, etc.)

IMHO this is likely to be over complex and error prone. I don't think we should try to be too smart here. I'd say just keep it simpler and properly document the behavior.

ipetrov117:
> This is what is fixing https://github.com/SUSE/elemental/pull/244

Sure, I do not argue that. If you like your approach better, I do not mind keeping the logic you introduce to fix the issue.

> I don't think this is an issue, but a feature. The caller should be aware of the layout it wants and the source of truth is the partitions order. If the caller wants to relabel system partition with ROOT label then this is possible, just keep empty (not nil) any of the previous partitions and merge will keep them and then at the system's partition position just add a partition only with the label defined. If you need to add a new partition somewhere in the middle then you need to provide the full layout, but still possible.

I am sorry, but I struggle to understand how is this supposed to work for the `customize` command when working with the ISO installer? With the current merge implementation `customize` is not able to merge partitions correctly and breaks the system quite hard. 

Lets say we have the following use-case (this example assumes customize now works with the config directory setup):
1. ISO image is built and inside it we have an `install.yaml` deployment file with the `EFI`, `RECOVERY` and `SYSTEM` partitions defined in it.
2. User uses this ISO image through the customize command, where the image comes from a release manifest.
3. User makes configurations of the image, by, let's say, adding Kubernetes and Network configuration to the configuration directory
4. This tells `customize` that there are 2 new partitions that need to be added - `ignition` and `ELEMENTAL-PREPARE`.

To mirror this theoretic use-case outside of the currently missing implementation (customize + config dir), let's say that `customize` will mirror the full partition setup and somewhere in the "middle" add its new partitions. So I provide a full layout (containing `EFI`, `ignition`, `ELEMENTAL-PREPARE`, `RECOVERY` and `SYSTEM`) in a `deployment.yaml` file and pass it to the current customize `--install-description`. And according to what you said, it should work, however what happens is that the partitions get mangled up and `ELEMENTAL-PREPARE` gets merged with `SYSTEM`, I do not see this as a feature, but as a bug. Maybe I am missing something.

>Yes, this is the intended behavior, order matters. There is no partition matching logic to keep merge logic simple, otherwise trying to be smart and matching partitions has high potential to hit corner cases because of the matching criteria (trying to relabel an already defined partition, changing UUIDS, partitions without a filesystem label defined, etc.)

Apologies, maybe this is ignorant, but to me, blindly merging partitions from one place to another and expecting that the user/process will always be aware of the full picture is weird.

Also, I tried to look at this from a user perspective. For example, as a user, if I want to add a new partition I do not expect to be intimately familiar with the full partition setup of the system and have to define it in my configuration just so that I can add my partition. Same goes for when editing a partition, I want to edit a specific partition and not have to define empty entries so that I can get to my partition. I guess I am missing a bigger picture here.

> IMHO this is likely to be over complex and error prone. I don't think we should try to be too smart here. I'd say just keep it simpler and properly document the behavior.

I really want to stress that I am not **trying to be smart**, or overcomplicate things, or undermine someone's work, far from it.. I simply was trying to fix something that does not seem to work. I am neither claiming that this is the best approach, nor saying that this is the way it should be done. I am perfectly happy with not merging this, but we need a stable merge command. The current implementation does not seem to be stable enough, as customize is struggling to correctly merge partitions over the ISO installer `install.yaml` file. Since this is not a good approach to you, can you please suggest a better one.

> Also, keep in mind the order of the final result also matters, this will be partitions order on the installed system. So in the suggested example the merged result is pushing the EFI partition to be the third one in the disk. How is the user controlling partitions position in the disk?

I agree that ordering can be improved. The current ordering adds first partitions from source then partitions from dest. I did this so that `SYSTEM` can be last, but I did not think that `EFI` needs to be earlier. This can for sure be improved, for example we can make sure that dest partition locations are preserved, where source partitions are added between all other dest partitions and `SYSTEM`. That way we ensure `EFI` and `RECOVERY` are not moved and `SYSTEM` is last. But from what I understand, this is not the biggest problem of the PR.

pkg/deployment/merge.go
ipetrov117: I may be missing something, but I think I am merging them by the Partition attribute [Label](https://github.com/ipetrov117/elemental/blob/merge-func-fix/pkg/deployment/merge.go#L66). I am not using a custom label.

Also I agree that merging by label is not ideal, however not all cases have UUIDs (i.e. there are no UUIDs in the `install.yaml` file in the ISO installer media), also Roles do not seem to be a good identifier as well.  I struggled to find a better identifier than partition label. Happy to hear suggestions.

> What if there are duplicated or crossed UUIDs?

Currently duplications are handled by taking the last element of the duplicates and merging only it. 

> And what if customize is actually trying to change the label?

This would be a limitation of the approach, due to not being able to find a better identifier. Happy to hear better alternatives.

> What if the label is undefined?

Partitions with undefined labels are just added to the final list.

> I am not a fan of identifying partitions by filesystem label, it tends to be not robust enough.

Agreed, what would be a better identifier in your opinion?

PR #269, opened 2025-12-02
Rework 'elemental3 customize' command to work with the configuration directory

ipetrov117:
This PR suggests an implementation that reworks the `elemental3 customize` command and enables it to work with the configuration directory.

Notable changes:
1. The `Core Platform` release manifest now contains the ISO installer container image as well
    - **Why?** - `elemental3 customize` will unpack this container image and use the ISO that was built under `/iso` as the base over which the customer will apply all desired customizations.
1. `diskSize` and `device` have been split under their own media specific configuration blocks
    - **Why?** - when building ISO images, `diskSize` is not supposed to be specified, as well as the other way around. This configuration approach takes inspiration from how EIB [used to define](https://github.com/suse-edge/edge-image-builder/blob/main/pkg/image/testdata/full-valid-example.yaml#L8-L11) media specific configuration types.
1. A new `customize.Runner` was introduced.
    - **Why?** - having the `customize.Runner` allows us to separate the customization logic out of the `action` package and have a clear distinction between customization initialisation (happening in `action`) and customization run (happening in `customize`). Furthermore this sets the ground for being able to actually unit test the customization run.
1. `elemental3 customize` now automatically applies an [installer live config script](https://github.com/ipetrov117/elemental/blob/customize-cmd-config-dir/internal/customize/templates/auto_installer.sh.tpl) that will do the installation for both RAW and ISO media types.
    - **Why?** - IIUC, having a [config-live.sh](https://src.suse.de/UnifiedCore/uc-factory/src/branch/main/OS/uc-base-os/config-live.sh) script defined as part of our container image builds was never the final solution. Having this close to the code makes more sense, as we can iterate easily over it and everything happens in one place.
1. Old flags for the `elemental3 customize` command have been removed.
    - **Why?** - many of the flags introduced complexities that we do not want to offer to the user from the start (e.g. defining a full installation description). We can revisit each of the flags when the time comes and if we find the need for them, they can easily be reintroduced.
1. `elemental3 customize` now exclusively runs with the `release manifest` logic.
    - **Why?** - to offer a fully robust image building solution that is based on a specific product that steps on top of our platform, we need to use a verified image from a release manifest. That said, it may be beneficial to also support a configuration where just an image is provided (e.g. either through a specific configuration in the config directory, or through a flag). The merit for this can further be explored and discussed after release.

How does the workflow of the `elemental3 customize` command look now?
1. User defines a configuration directory, by following our [docs](https://github.com/SUSE/elemental/blob/main/docs/configuration-directory.md) (when we get them in shape, of course). 
1. User runs `elemental3 customize --type <raw/iso> --config-dir <path> [optional params]`
1. Elemental:
   - Locates the `uc-base-kernel-default-iso` container image defined in the `Core Platform` manifest that was either specified by the user, or specified by a product release manifest that the user was referring to.
   - Unpacks said container image and extracts the ISO installer media, located at `/iso`.
   - Configures the user specified components under the `overlays` directory.
   - Prepares a customization `deployment.Deployment` object.
   - Runs the customization process using said object.
1. End result is either a `.raw` or `.iso` media under the specified output directory.
1. When booting the image, the installer will automatically run and setup the system without any user intervention needed.

The below tests have been run for both `RAW` and `ISO` media types:
1. Linux-only build (e.g. no Kubernetes).
2. Single node Kubernetes cluster.
3. Multi-node Kubernetes cluster (3 control-plane + 1 worker).

Documentation PR will follow once we agree whether the proposed implementation makes sense.

### DIFF
--- a/internal/cli/action/build.go
+++ b/internal/cli/action/build.go
@@ -18,7 +18,6 @@ limitations under the License.
 package action
 
 import (
-	"errors"
 	"fmt"
 	"os/signal"
 	"path/filepath"
@@ -39,7 +38,7 @@ import (
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
 )
 
-func Build(ctx *cli.Context) (err error) {
+func Build(ctx *cli.Context) error {
 	args := &cmd.BuildArgs
 
 	if ctx.App.Metadata == nil || ctx.App.Metadata["system"] == nil {
@@ -77,8 +76,7 @@ func Build(ctx *cli.Context) (err error) {
 		logger.Debug("Cleaning up build-dir %s", outDir)
 		rmErr := system.FS().RemoveAll(string(outDir))
 		if rmErr != nil {
-			logger.Error("Cleaning up build-dir %s", outDir)
-			err = errors.Join(err, rmErr)
+			logger.Error("Cleaning up build-dir '%s' failed: %v", outDir, rmErr)
 		}
 	}()
 
--- a/internal/cli/action/customize.go
+++ b/internal/cli/action/customize.go
@@ -19,7 +19,6 @@ package action
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os/signal"
 	"path/filepath"
@@ -40,7 +39,7 @@ import (
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
 )
 
-func Customize(ctx *cli.Context) (err error) {
+func Customize(ctx *cli.Context) error {
 	if ctx.App.Metadata == nil || ctx.App.Metadata["system"] == nil {
 		return fmt.Errorf("error setting up initial configuration")
 	}
@@ -64,8 +63,7 @@ func Customize(ctx *cli.Context) (err error) {
 		logger.Debug("Cleaning up customize-dir %s", outDir)
 		rmErr := system.FS().RemoveAll(string(outDir))
 		if rmErr != nil {
-			logger.Error("Cleaning up customize-dir '%s' failed", outDir)
-			err = errors.Join(err, rmErr)
+			logger.Error("Cleaning up customize-dir '%s' failed: %v", outDir, rmErr)
 		}
 	}()
 


internal/cli/action/customize.go
```diff
@@ -18,97 +18,144 @@ limitations under the License.
...
 	defer cancelFunc()
 
 	logger.Info("Customizing image")
 
-	mType, err := installer.StringToMediaType(args.Type)
+	customizeDir := fmt.Sprintf("customize-%s", time.Now().UTC().Format("2006-01-02T15-04-05"))
+	outDir, err := config.CreateOutputDir(system.FS(), args.CustomizeOutput, customizeDir, 0700)
 	if err != nil {
-		return fmt.Errorf("invalid media type for customize: %w", err)
+		logger.Error("Creating customize directory failed")
```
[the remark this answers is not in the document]
ipetrov117: The error will be returned at the top level, but I agree it may be a more descriptive to at least specify the directory in question. Fixed.

internal/cli/action/build.go
```diff
@@ -65,17 +66,19 @@ func Build(ctx *cli.Context) error {
...
 	}
 
 	defer func() {
 		logger.Debug("Cleaning up build-dir %s", outDir)
-		err = system.FS().RemoveAll(string(outDir))
-		if err != nil {
+		rmErr := system.FS().RemoveAll(string(outDir))
+		if rmErr != nil {
 			logger.Error("Cleaning up build-dir %s", outDir)
+			err = errors.Join(err, rmErr)
```
atanasdinov: This is probably an overkill. Even if this piece fails, the overall "build" process should be successful if `err` is nil here.

internal/cli/action/customize.go
```diff
@@ -18,97 +18,146 @@ limitations under the License.
...
+			logger.Error("Cleaning up customize-dir %s", outDir)
+			err = errors.Join(err, rmErr)
+		}
+	}()
 
-	d, err := digestCustomizeDeploymentSetup(s, args)
+	def, err := digestCustomizeDefinition(system.FS(), args)
 	if err != nil {
-		s.Logger().Error("Failed to collect deployment setup")
+		logger.Error("Digesting image definition from customize flags")
```
atanasdinov:
```suggestion
		logger.Error("Digesting image definition from customize flags failed")
```

Many of the other `.Error()` logs also don't contain "failed" or "error" and it reads like an info log but on the error level. Could we fix these too?

atanasdinov (commented):
Looks great with what our current options. We'll simplify (or completely get rid of) the merge functionality in the future.

internal/customize/customize_test.go
```diff
@@ -0,0 +1,431 @@
...
+			},
+			Media: &mediaMock{
+				customizeFunc: func(d *deployment.Deployment) error {
+					return nil
+				},
+			},
+		}
+		sideEffects["xorriso"] = func(args ...string) ([]byte, error) {
+			file := filepath.Join(string(outputDir), "iso-desc-install", "install.yaml")
+			Expect(fs.WriteFile(file, []byte(installerDescr), vfs.FilePerm)).To(Succeed())
```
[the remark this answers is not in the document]
ipetrov117: My main reasoning for creating the file in the `xorriso` mock was that it clearly shows what we are actually mocking and why we are mocking it. In the [original logic ](https://github.com/SUSE/elemental/blob/main/pkg/installer/media.go#L153) we have a call to the [extractISO](https://github.com/SUSE/elemental/blob/main/pkg/installer/media.go#L140) function, which uses `xorriso` to extract the file to the image. Creating the file as part of the `xorriso` mock made more sense than just creating the file as part of the test FS, as it clearly showed the reasoning behind this mock.

Let me know if this makes sense to you.

internal/customize/customize.go
```diff
@@ -0,0 +1,310 @@
...
+	// preparedPartitions = [{}, {}, nil, add..., SYSTEM].
+	//
+	// When this result is merged with the installer partition slice - '[EFI, RECOVERY, SYSTEM]';
+	// we will get the following merged result:
+	// mergedPartitions = [EFI, RECOVERY, add..., SYSTEM]
+	return append(preparedPartitionSlice, src[len(src)-1])
+}
+
+func resizeDisk(runner sys.Runner, disk string, diskSize install.DiskSize) error {
+	const defaultSize = "10G"
```
davidcassany: nit: I wonder if we could skip resize if no size is provided. Basically the generated disk will auto expand as soon as it is dumped to a bigger disk, which I think it is the primary use case as this limits the amount of free space in the generated image and makes it easier to handle and transport.

internal/customize/customize.go
```diff
@@ -0,0 +1,310 @@
...
+
+	if diskSize == "" {
+		diskSize = defaultSize
+	} else if !diskSize.IsValid() {
+		return fmt.Errorf("invalid disk size definition '%s'", diskSize)
+	}
+
+	_, err := runner.Run("truncate", "-s", string(diskSize), disk)
```
davidcassany: nit: I'd rather use os.File.Truncate which is then supporting the vfs interface and you could eventually handle the spare vs non spare case by writing some bytes at the end or not.

PR #272, opened 2025-12-04
Set kubernetes config file paths

frelon:
When the manifests dir does not exist we still want to check if
{server,agent}.yaml kubernetes config files exist and set accordingly.


ldevulder (approved):
Validated on my lab, LGTM!

I just saw lint issues that are not related to your PR, I don't know why we have those now.

atanasdinov (approved):

PR #278, opened 2025-12-04
Simplify release manifest schema

atanasdinov:
This change simplifies the release manifest schema:
- all unused and unnecessary properties are dropped
- all image based values now follow the same schema "registry-url:tag"


dirkmueller (approved):
I like it.

PR #286, opened 2025-12-05
Only run k8s-resource-installer on init-node

frelon:
This commit only runs the k8s-resource-installer service on the
configured init-node by setting ConditionHost= on the systemd unit.

If no nodes are specified (single-node deploy), the service will run for
any hostname.


ldevulder (approved):
LGTM, thanks for the quick fix!

davidcassany (approved):

PR #293, opened 2025-12-09
Improve UUIDs assignement after running systemd-repart

davidcassany:
This PR ensures systemd-repart reported UUIDs are assigned to the defined partitions in Deployment structure. `systemd-repart` lists new disk layout on success, which is an array of partitions. Any partition in the disk that was not included in the given configuration files is listed with the `file` attribute set to  `null`, otherwise `file` refers to the configuration file this partition was paired with.

### DIFF
--- a/pkg/repart/disk_repart.go
+++ b/pkg/repart/disk_repart.go
@@ -228,6 +228,7 @@ func runSystemdRepart(s *sys.System, target string, parts []Partition, flags ...
 		File string `json:"file,omitempty"`
 	}{}
 
+	s.Logger().Debug("systemd-repart output to parse:\n%s", string(out))
 	err = json.Unmarshal(out, &uuids)
 	if err != nil {
 		return fmt.Errorf("failed parsing systemd-repart JSON output: %w", err)


pkg/repart/disk_repart.go
```diff
@@ -221,20 +224,26 @@ func runSystemdRepart(s *sys.System, target string, parts []Partition, flags ...
...
+		UUID string `json:"uuid,omitempty"`
+		File string `json:"file,omitempty"`
 	}{}
 
 	err = json.Unmarshal(out, &uuids)
 	if err != nil {
 		return fmt.Errorf("failed parsing systemd-repart JSON output: %w", err)
 	}
-	if len(uuids) != len(parts) {
-		return fmt.Errorf("partitions mismatch between Deployment and systemd-repart JSON output: %s", string(out))
```
atanasdinov: I'd suggest to always log the output on a debug level above. It proves to be really important to understand the sequence.

atanasdinov (approved):

ipetrov117 (approved):

PR #294, opened 2025-12-09
WIP: append config-2 partition to customized RAW images

davidcassany:


davidcassany:
Closing the PR as this was only a rough draft to actually test it with SUSE Edge IPA. Whenever we take action again on the topic we should restart it from scratch, this was just a simple ugly attempt to validate if it was possible to prevent SUSE Edge IPA to create a config drive.

PR #297, opened 2025-12-10
Fix k8s-resource-installer execution on init node

ldevulder:
In DHCP the `ConditionHost` recently added is not working because the service is started too quickly. This is due to a misconfiguration of `After` condition, as the real service started is not `rke2-server` but `k8s-config-installer` instead.

### DIFF
--- a/internal/config/templates/k8s-resource-installer.service.tpl
+++ b/internal/config/templates/k8s-resource-installer.service.tpl
@@ -1,6 +1,6 @@
 [Unit]
 Description=Kubernetes Resources Installer
-After=rke2-server.service
+After=k8s-config-installer.service
 ConditionHost={{ .InitHostname }}
 
 [Service]


internal/config/templates/k8s_conf_deploy.sh.tpl
```diff
@@ -8,30 +8,28 @@ declare -A hosts
 hosts[{{ .Hostname }}]={{ .Type }}
 {{- end }}
 
-HOSTNAME=$(cat /etc/hostname)
-if [ ! "$HOSTNAME" ]; then
-    HOSTNAME=$(cat /proc/sys/kernel/hostname)
-fi
-
-NODETYPE="${hosts[$HOSTNAME]:-server}"
+HOSTNAME=$(</proc/sys/kernel/hostname)
```
ldevulder: The hostname seems to be always defined in `/proc/sys`.

dirkmueller: is this actually fixing something or was this unrelated cleanup? I do agree that /proc/sys/kernel/hostname should be preferred, but I just feel nervous about doing that change without extensive testing. 

if it was unrelated can we just please drop it and push it to another PR? I'm happy with all the other changes.

dirkmueller (approved):

ipetrov117 (approved):

PR #302, opened 2025-12-11
Fix creation of checksums in customize workflow

dirkmueller:


internal/image/install/install.go
```diff
@@ -18,8 +18,13 @@ limitations under the License.
 package install
 
 import (
+	"math/bits"
 	"regexp"
+	"strconv"
+	"strings"
+	"unicode"
 
+	"github.com/docker/go-units"
```
davidcassany: nit: shouldn't this become and explicit requirement in go.mod?

pkg/installer/media.go
```diff
@@ -93,12 +94,13 @@ type Media struct {
...
-	unpackOpts []unpack.Opt
-	bl         bootloader.Bootloader
-	outputFile string
+	mType       MediaType
+	s           *sys.System
+	ctx         context.Context
+	unpackOpts  []unpack.Opt
+	bl          bootloader.Bootloader
+	outputFile  string
+	rawDiskSize imginstall.DiskSize
```
davidcassany: nit: Wouldn't be simpler running the convert and check in advance and simply store the required MiB in here?

davidcassany (approved):
Nice! Just a couple of small comments. Feel free to ignore them or leave it for later as it can be rearranged when addressing the TODO.

ipetrov117 (approved):

PR #308, opened 2025-12-11
Introduce customisation mode option

atanasdinov:
Introduces a new `--mode` CLI parameter to customize command with the following options:
- embedded (default): Customising an image will embed the Ignition partition within the resulting image
- split: Customising an image will provide the ready-to-install image and Ignition configuration directory separately

The use cases that are going to utilise this feature are:
- CAPI: Base image only without any configuration embedded
- Any environment where having the configuration separate is desired
- Baseline image testing: Customise an image only once and generate various Ignition configurations only on demand (once we add in `--mode config-only`)

In order to utilise the configuration directory in a QEMU VM, you can simply:
```
$ mkisofs -o ignition.iso -V ignition <config-dir>
$ qemu-kvm ... -cdrom ignition.iso
```

Note: Currently the networking, Ignition data, and custom scripts are going to be extracted to the configuration directory. All files which are part of the overlay (incl. systemd extensions, Kubernetes manifests, Helm charts) will proceed to be part of the image. We will have to properly distribute those to the configuration directory instead later.

### DIFF
--- a/internal/cli/cmd/customize.go
+++ b/internal/cli/cmd/customize.go
@@ -64,7 +64,7 @@ func NewCustomizeCommand(appName string, action func(*cli.Context) error) *cli.C
 			},
 			&cli.StringFlag{
 				Name:        "mode",
-				Usage:       "Customization mode, 'embedded' or 'split'",
+				Usage:       "Customization mode, 'embedded' (config in image) or 'split' (config separate)",
 				Destination: &CustomizeArgs.Mode,
 				Value:       "embedded",
 			},


ipetrov117 (approved):

internal/cli/action/customize.go
```diff
@@ -50,76 +51,95 @@ func Customize(ctx *cli.Context) error {
...
+	if err = customizeRunner.Run(ctxCancel, def, output, args.Local); err != nil {
 		logger.Error("Customizing installer media failed")
 		return err
 	}
 
 	return nil
 }
 
+func resolveOutputPaths(args *cmd.CustomizeFlags) (imagePath, configPath string) {
+	imagePath = args.OutputPath
```
dharmit: Would a `nil` check be worth it here?

atanasdinov: I don't seem to understand -- the args variable cannot be nil no matter what we do.

internal/cli/cmd/customize.go
```diff
@@ -61,6 +62,12 @@ func NewCustomizeCommand(appName string, action func(*cli.Context) error) *cli.C
 				Destination: &CustomizeArgs.OutputPath,
 				DefaultText: "image-<timestamp>.<image-type>",
 			},
+			&cli.StringFlag{
+				Name:        "mode",
+				Usage:       "Customization mode, 'embedded' or 'split'",
```
dirkmueller: It is quite hard to understand for me. maybe

```go

   Usage:  "configuration output mode, 'embedded' (config within image, default) or 'split' (config separate to image)" 
```

dirkmueller (approved):

PR #310, opened 2025-12-12
Predefine output file for media

atanasdinov:
- Customising images now accepts the user provided `--output` value or the generated `image-<timestamp>.<type>` one
- Validates that --mode option can only be `embedded` or `split`


atanasdinov:
This line references how the default output would look and it's correct.

ipetrov117 (approved):

PR #311, opened 2025-12-12
also support --output pointing to a directory

dirkmueller:
otherwise --output path/to/dir produces `path/to/dir.raw`. with this change it produces `path/to/dir/image-$timestamp.raw`.

### DIFF
--- a/internal/cli/action/customize.go
+++ b/internal/cli/action/customize.go
@@ -50,9 +50,9 @@ func Customize(ctx context.Context, cmd *cli.Command) error {
 	fs := system.FS()
 	args := &cmdpkg.CustomizeArgs
 
-	logger.Info("Customizing image started")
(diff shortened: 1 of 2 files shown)


dirkmueller:
> I was referring to this being the entry point:
> 
> > otherwise --output path/to/dir produces path/to/dir.raw


True, I wrote this description and test and code when f13b89413c18bfb51d669eb417cbdfe92a0f40df was not yet merged in the tree. To reproduce, you would have to revert to a state earlier than that. with the change above the behavior changed slightly. I still think we shouldn't error out on --output pointing to an existing directory and with this change it will just create an image wihtin the directory, which is a sensible behavior imho. 

It is more of a new feature then though.

frelon (approved):

davidcassany (approved):

PR #320, opened 2025-12-16
Add toolbox documentation (jsc#CORE-414)

dirkmueller:


docs/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
+# Troubleshooting an installed system
+
+
+## Using `toolbox` for debugging
+
+Operating Systems deployed by Elemenal are image based and immutable during
+runtime. Any change would be applied only after reboot. That solution has several benefits, but it also has some disadvantages. If you need to debug your system and install a new tool, the tool will be available only after reboot. Therefore, you cannot debug the currently running system. For this reason, a utility called toolbox has been developed.
```
atanasdinov: > Any change would be applied only after reboot.

> If you need to debug your system and install a new tool, the tool will be available only after reboot.

Is this copied from Micro docs? None of it applies to Elemental.

dirkmueller: Removed.

README.md
```diff
@@ -23,6 +23,8 @@ Elemental is a tool for installing, configuring and updating operating system im
 * [Image Customization](./docs/image-customization.md) - for users and/or consumers interested in customizing images that are based on a specific release.
 * [Release Manifest Guide](./docs/release-manifest.md) - for consumers interested in creating a release manifest for their product.
 * [Elemental and Ignition Integration](./docs/ignition-integration.md) - for consumers interested in understanding the nuances and capabilities of Ignition in the scope of Elemental.
+* [Troubleshooting Guide](./docs/troubleshooting.md) - guide for users and
+ consumers in troubleshooting
```
atanasdinov: Feel free to use any other wording but the majority of the documentation that we have is about customising bootable artefacts. If we don't specify that this guide is for live / running hosts, then it could lead to some confusion.
```suggestion
 consumers in troubleshooting running systems
```

docs/troubleshooting.md
```diff
@@ -0,0 +1,34 @@
...
+
+```shell
+# podman rm toolbox-root
+```
+
+## Using `toolbox`
+
+In the toolbox container, you can install any tool you want with zypper and then use the tool without rebooting your system. It does not require the system to be registered, as a subset of SUSE Linux Enterprise Server packages are available for installation.
+
+To leave the container, just type exit. Remember that the container stays in the same state as you exit it. If you want a clean environment, you need to remove the toolbox container first.
```
atanasdinov:
```suggestion
To leave the container, just type `exit`. Remember that the container stays in the same state as you exit it. If you want a clean environment, you need to remove the toolbox container first.
```

dirkmueller:
> Absolute nits.

addressed, thank you!

PR #326, opened 2026-01-06
Miscellaneous changes based on `elemental3 customize`

dharmit:
I was running `elemental3 customize` on local setup to get a hang of the flow of things. This PR includes miscellaneous changes like typos, variable removal, declaring variable differently, docs changes, etc.

There's no pattern to these changes and some might not make sense, which can be undone.

### DIFF
--- a/docs/configuration-directory.md
+++ b/docs/configuration-directory.md
@@ -1,6 +1,6 @@
 # Configuration Directory Guide
 
-The configuration directory is the place where users define the desired state of the image they intend to build by using the [elemental3 build](image-build-and-customization.md#elemental3-build) command.
+The configuration directory is the place where users define the desired state of the image they intend to build by using the [elemental3 customize](image-customization.md) command.
 
 Generally, the available configuration areas that this directory supports are the following:
 
--- a/docs/image-customization.md
+++ b/docs/image-customization.md
@@ -98,11 +98,19 @@ Run the customization process:
 >
 > 1. Start Podman socket:
 >   ```shell
->   systemctl enable --now podman.socket
+>   sudo systemctl enable --now podman.socket
+>   ```
+>   or, for rootless containers with Podman:
+>   ```shell
+>   systemctl --user enable --now podman.socket
 >   ```
 > 2. Run the `elemental3` container with the mounted podman socket:
 >   ```shell
->   podman run -it -v <PATH_TO_CONFIG_DIR>:/config -v /run/podman/podman.sock:/var/run/docker.sock $ELEMENTAL_IMAGE customize --type <raw/iso> --local
+>   sudo podman run -it -v <PATH_TO_CONFIG_DIR>:/config -v /run/podman/podman.sock:/var/run/docker.sock $ELEMENTAL_IMAGE customize --type <raw/iso> --local
+>   ```
+>   or, if using rootless containers with Podman:
+>   ```shell
+>   podman run -it -v <PATH_TO_CONFIG_DIR>:/config -v /run/$XDG_RUNTIME_DIR/podman.sock:/var/run/docker.sock $ELEMENTAL_IMAGE customize --type <raw/iso> --local
 >   ```
 
 Unless configured otherwise, the above process will produce a customized RAW or ISO image under the specified `<PATH_TO_CONFIG_DIR>` directory.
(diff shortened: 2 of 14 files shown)


internal/config/kubernetes.go
```diff
@@ -59,14 +59,15 @@ func (m *Manager) configureKubernetes(
 	conf *image.Configuration,
 	manifest *resolver.ResolvedManifest,
 	output Output,
-) (k8sResourceScript, k8sConfScript string, err error) {
+) (string, string, error) {
```
atanasdinov: Is there a benefit to this? Naming the returned variables clearly indicate what their contents are.

docs/image-customization.md
```diff
@@ -151,7 +159,7 @@ The customized image can be booted as any other regular image. Below you can fin
                   --ram 16000 \
                   --vcpus 10 \
                   --import \
-                  --disk path=disk.img,format=raw \
+                  --disk path=disk.img,format=iso \
```
atanasdinov: The disk path `disk.img` is a RAW file.

docs/configuration-directory.md
```diff
@@ -1,6 +1,6 @@
 # Configuration Directory Guide
 
-The configuration directory is the place where users define the desired state of the image they intend to build by using the [elemental3 build](image-build-and-customization.md#elemental3-build) command.
+The configuration directory is the place where users define the desired state of the image they intend to build by using the [elemental3 customize](image-customization.md) command.
```
atanasdinov:
```suggestion
The configuration directory is the place where users define the desired state of the bootable image, following the [customization process](image-customization.md).
```

atanasdinov (changes requested):
There are some pieces that we'd have to adjust before merging, I've commented on those below.

Additionally, trying to unite numerous different changes over code, documentation and examples in a single "misc" commit is not a great practice. We are following a rebasing strategy where we prefer to keep the history in a clean state in order to easily track what has changed over the course of time.

Thanks for bringing in these improvements!

pkg/manifest/api/core/manifest.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/suse/elemental/v3/pkg/manifest/api"
 )
 
+// ReleaseManifest TODO (dharmit): why not call this CoreManifest?
```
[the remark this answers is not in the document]
atanasdinov: Both are perfectly valid -- there are core and product release manifests.

internal/cli/cmd/customize.go
```diff
@@ -82,7 +81,7 @@ func NewCustomizeCommand(appName string, action func(*cli.Context) error) *cli.C
 				Name:        "platform",
 				Usage:       "Target platform",
 				Destination: &CustomizeArgs.Platform,
-				Value:       fmt.Sprintf("linux/%s", runtime.GOARCH),
+				Value:       "linux/amd64",
```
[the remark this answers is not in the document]
dharmit: Aren't we supporting amd64 platform only? My understanding is that, by default, we build artifacts for amd64 only.

That's why I suggested this change, as running elemental on different platforms would generate artifact for the platform it's running on, which is obviously nice, but I thought (or, assumed) we want to stick to what we support.

Anyway, I'll revert.

internal/config/kubernetes.go
dharmit: More return statements in that function use `""` for the two string parameters than named return variables. Plus, a bit of my bias toward not using named return values, as they make code less readable if a naked return was used.

internal/cli/cmd/customize.go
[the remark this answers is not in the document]
atanasdinov: Support for both amd64 and arm64 is necessary. We might not be testing arm64 yet, but we shouldn't introduce such invasive change that would flat out break it unless the argument is supplied.

atanasdinov:
> Confused. How would a single commit modifying numerous different files prevent us from keeping the history clean? Asking to learn because there's clearly something I don't understand. And I actually like the idea about keeping history clean. But, tbh, I don't know what bad git history and good git history look like.

Using the changes in this PR, a good practice would be something along the lines of:
- Fix docs link
- Append rootless Podman examples
- Fix code comments
- Adjust string quotation logs

Intentionally continuing the examples with a change I asked to be reverted:
- Use linux/amd64 as a default value for --platform argument
   - It is expected that only x86_64 hosts would be the supported target platforms for the foreseeable future. Overriding the argument is still possible but not guaranteed to work.

Now, if all of the above gets done in a single commit "Misc changes", we've lost all the context of what has happened. Someone looking at the `--platform` argument implementation would see a "misc change" without any sort of intention and be lost as to why that happened.

There are various sources and opinions on the Internet about what a proper commit is. Feel free to look it up. A lot of those would say things similar to:
- Each commit should be as short as possible
- Each commit should ideally describe a single change
- Each commit should use a longer description including additional context if the change itself does not carry enough information as to why it is done

PR #331, opened 2026-01-12
Integration tests for "elemental3 customize"

ipetrov117:
This PR suggests an implementation for integration tests that relate to the `elemental3 customize` command.

About the integration tests themselves:
1. _What is the use-case that is being tested?_ - as discussed, the tests are handling the [linux-only image](https://github.com/SUSE/elemental/blob/main/docs/image-customization.md#linux-only-image) use-case with the following customizations enabled:
   - Statically defined network.
   - Custom systemd service.
   - Custom kernel cmdline.
   - FIPS enabled.
1. _What are the actual tests?_ - tests are fairly simple and cover the following:
   - Are we running the expected OS?
   - Are we running the expected network?
      - Has the hostname been defined correctly as part of the static network definition?
      - Check various network properties (e.g. is it up, is it of the expected type, does it hold the expected IP, etc..)
   - Are we running the OS with FIPS enabled?
   - Has the custom kernel cmdline been propagated?
   - Has the custom systemd service been correctly setup?
1. _How do the tests affect the current workflow?_ - test are run after the existing `elemental3ctl install` tests. Between each test we do a cleanup to ensure that the `run_vm.sh` script can successfully build the next VM. The alternative to this was doing a cleanup after all tests, but this proved to be less effective and readable, as multiple needless properties and configurations had to be made so that `run_vm.sh` could boot 3 VMs next to one another.

_What is left?_ - Currently these tests are running with a release manifest that is taken directly from `devel:UnifiedCore:Tumbleweed`. This needs to change, so that the tests take a release manifest from the respective OBS project that was created for the respective PR. To do this we would need to first land #319. After that the only change to this PR would be a reference to the respective release manifest, which should be a minimal change. Nevertheless, until that happens this PR will stay in **Draft**.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -13,7 +13,7 @@ OS_VERSION?=latest
 DOCKER_SOCK?=/var/run/docker.sock
 
 ### Elemental customize specific variables ###
-RELEASE_MANIFEST?=registry.opensuse.org/devel/unifiedcore/tumbleweed/containers/release-manifest:latest
+RELEASE_MANIFEST?=$(IMG_REPO)/release-manifest:latest
 CONFIG_DIR?=$(abspath ./tests/testdata/config-dir)
 CUSTOMIZED_WORKDIR_NAME=customized
 CUSTOMIZED_WORKDIR_PATH?=$(CONFIG_DIR)/$(CUSTOMIZED_WORKDIR_NAME)


tests/testdata/config-dir/butane.yaml
```diff
@@ -0,0 +1,40 @@
...
+        [Install]
+        WantedBy=multi-user.target
+storage:
+  files:
+    - path: /etc/ssh/sshd_config.d/root_access.conf
+      mode: 0644
+      contents:
+        inline: |
+          PermitRootLogin yes
+    - path: /var/lib/elemental/example/example.sh
```
dirkmueller: should we put this into /usr/local/bin perhaps?

PR #361, opened 2026-02-13
Add kernel cmdline only once

### LINKED ISSUE #299: Options appear twice in Grub configuration
While booting a FIPS enabled OS image I saw that some options are set twice (`fips` and `boot=` here):
Image

dharmit:
Fixes the issue of `fips=1` and `boot=LABEL=EFI` appearing twice in kernel cmdline after installation.

Fixes #299.

### DIFF
--- a/pkg/fips/fips.go
+++ b/pkg/fips/fips.go
@@ -20,6 +20,7 @@ package fips
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/suse/elemental/v3/pkg/chroot"
 	"github.com/suse/elemental/v3/pkg/deployment"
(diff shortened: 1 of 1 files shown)


jtomasek (approved):

internal/customize/customize_test.go
```diff
@@ -423,8 +423,8 @@ func (m *mediaMock) Customize(d *deployment.Deployment) error {
 func defaultCustomizeDeploymentValidation(dep *deployment.Deployment) {
 	Expect(dep.BootConfig.Bootloader).To(Equal("grub"))
 
-	expectedCMD := fmt.Sprintf("console=ttyS0 %s %s", "fips=1", fmt.Sprintf("boot=LABEL=%s", deployment.EfiLabel))
-	Expect(dep.BootConfig.KernelCmdline).To(Equal(expectedCMD))
+	Expect(dep.BootConfig.KernelCmdline).To(ContainSubstring("fips=1"))
```
ipetrov117: This seems to alter what we are actually validating. The deleted line validates that the full kernel command line matches something, while the suggested changes only checks if fips is present in the string. Meaning that if something else is added to the kernel command line by accident, the validation will not notice it.

To correctly tests the contents of the expected kernel command line, can't we just do:

```golang
expectedCMD := fmt.Sprintf("console=ttyS0 %s %s", fmt.Sprintf("boot=LABEL=%s", deployment.EfiLabel), "fips=1")
```

ipetrov117 (approved):

davidcassany (approved):

PR #387, opened 2026-03-17
Allow users to provide credentials for Helm Charts

dbw7:
Introduces the ability for users to provide credentials for Helm charts so they can be pulled at build time

Example usage can be seen here:
`release.yaml`
```
helm:
  - chart: endpoint-copier-operator
    credentials:
      username: USER
      password: PASS
```
`cluster.yaml`
```
helm:
  charts:
    - name: "suse-storage"
      version: "1.9.2"
      targetNamespace: "longhorn-system"
      repositoryName: "appco"
      valuesFile: "longhorn.yaml"
    - name: "second-chart"
      version: 1.0.0
      repositoryName: "appco"
  repositories:
    - name: "appco"
      url: "oci://dp.apps.rancher.io/charts"
      credentials:
        username: USER
        password: PASS
```

### DIFF
--- a/internal/config/kubernetes.go
+++ b/internal/config/kubernetes.go
@@ -41,8 +41,8 @@ var k8sResDeployScriptTpl string
 //go:embed templates/k8s_conf_deploy.sh.tpl
 var k8sConfDeployScriptTpl string
 
-func needsManifestsSetup(conf *image.Configuration) bool {
-	return len(conf.Kubernetes.RemoteManifests) > 0 || len(conf.Kubernetes.LocalManifests) > 0 || conf.Kubernetes.Network.IsHA()
+func needsManifestsSetup(conf *image.Configuration, additionalManifests map[string][]byte) bool {
+	return len(conf.Kubernetes.RemoteManifests) > 0 || len(conf.Kubernetes.LocalManifests) > 0 || conf.Kubernetes.Network.IsHA() || additionalManifests != nil
 }
 
 func needsHelmChartsSetup(conf *image.Configuration) bool {
@@ -50,7 +50,7 @@ func needsHelmChartsSetup(conf *image.Configuration) bool {
 }
 
 func isKubernetesEnabled(conf *image.Configuration) bool {
-	return conf.Release.Components.Kubernetes != nil || needsHelmChartsSetup(conf) || needsManifestsSetup(conf)
+	return conf.Release.Components.Kubernetes != nil || needsHelmChartsSetup(conf) || needsManifestsSetup(conf, nil)
 }
 
 func (m *Manager) configureKubernetes(
@@ -68,20 +68,21 @@ func (m *Manager) configureKubernetes(
 	}
 
 	var runtimeHelmCharts []string
+	var additionalManifests map[string][]byte
 	if needsHelmChartsSetup(conf) {
 		m.system.Logger().Info("Configuring Helm charts")
 
-		runtimeHelmCharts, err = m.helm.Configure(conf, manifest)
+		runtimeHelmCharts, additionalManifests, err = m.helm.Configure(conf, manifest)
 		if err != nil {
 			return "", "", fmt.Errorf("configuring helm charts: %w", err)
 		}
 	}
 
 	var runtimeManifestsDir string
-	if needsManifestsSetup(conf) {
+	if needsManifestsSetup(conf, additionalManifests) {
 		m.system.Logger().Info("Configuring Kubernetes manifests")
 
-		runtimeManifestsDir, err = m.setupManifests(ctx, &conf.Kubernetes, output)
+		runtimeManifestsDir, err = m.setupManifests(ctx, &conf.Kubernetes, additionalManifests, output)
 		if err != nil {
 			return "", "", fmt.Errorf("configuring kubernetes manifests: %w", err)
 		}
@@ -107,7 +108,7 @@ func (m *Manager) configureKubernetes(
 	return k8sResourceScript, k8sConfScript, nil
 }
 
-func (m *Manager) setupManifests(ctx context.Context, k *kubernetes.Kubernetes, output Output) (string, error) {
+func (m *Manager) setupManifests(ctx context.Context, k *kubernetes.Kubernetes, additionalManifests map[string][]byte, output Output) (string, error) {
 	fs := m.system.FS()
 
 	relativeManifestsPath := filepath.Join("/", image.KubernetesManifestsPath())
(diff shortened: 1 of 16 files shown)


ipetrov117 (commented):
Thanks for your efforts on this 🚀 

I left out some comments on things I am not entirely clear on. Let me know what you think!

frelon (approved):

internal/image/kubernetes/kubernetes.go
```diff
@@ -90,13 +90,19 @@ func (c *HelmChart) GetRepositoryName() string {
...
+func (c *HelmChart) ToCRD(values []byte, repository string, hasAuth bool) *helm.CRD {
+	return helm.NewCRD(c.TargetNamespace, c.Name, c.Version, string(values), repository, hasAuth)
 }
 
 type HelmRepository struct {
-	Name string `yaml:"name" validate:"required"`
-	URL  string `yaml:"url" validate:"required,url"`
+	Name        string      `yaml:"name" validate:"required"`
+	URL         string      `yaml:"url" validate:"required,url"`
+	Credentials Credentials `yaml:"credentials,omitempty"`
```
ipetrov117: Instead of defining a new struct, can't we just pass the existing [auth.Credentials](https://github.com/dbw7/elemental/blob/auth/internal/image/auth/auth.go#L29-L32) struct?

dirkmueller:
> Let's take the "single isolated credentials file" (e.g. `secrets.yaml/auth.yaml`) thesis for example. If we go down this route, the "editing multiple yaml files all over the place" will come from the fact that for each private resource, the user would have to jump between:
> 
> 1. `kubernetes/cluster.yaml` - for defining local Helm repositories
> 2. `release.yaml` - for defining Helm charts coming from the release manifest
> 3. The release manifest itself - for retrieving the necessary URL to which this credentials map to
> 4. The `secrets.yaml / auth.yaml` unified credentials file - for actually defining the credentials

I am not sure I get that. The core and product release manifest define repositories, each with a name: 

```
    repositories:
    - name: metallb
      url: https://metallb.github.io/metallb
    - name: suse-edge
      url: https://suse-edge.github.io/charts
```

The user doesn't have to modify anything there. the only thing they have to do is in the secrets.yaml to define their secrets for the indivudual repositories. example below:

```
cat secrets.yaml
  secrets:
  - name: metallb
    user: ..
    password: ...
  - name: suse-edge
    user: ..
    password: ..
  - name: rancher-charts-secret
    user: ..
    password: ..
```

No modification are needed on release manifest, because if we know these are behind access control, they would already be defining a "secret: <metallb>" link for example. we could even decide to make it implicitly linked via the repository name. if a chart or resource repository does not require authentication, that would be optional and no extra work needed.


> From my PoV this will be very prone to user error, as this involves jumping through multiple files **during configuration definition**. Furthermore, we should also not disregard the fact that we will be introducing a new configuration file. Our users struggle with the number of files as it is today, adding another one will just pile onto that frustration.

If that's a concern, we can define the "secrets" as part of "release.yaml" also. My main point is to not sprinkle secrets across many files, but centralize it in one location. having the file separate means we can easily remove it from "exports" for supportability purposes. The other advantage is that users who use private infrastructure can define their "private-registry" secret and use that everywhere, not falling out of sync (following DRY principle).

> Now on the other side of the coin is "embedding the credentials next to the resource definition" thesis. Here, as you already mentioned, the "editing multiple yaml files all over the place" will come when users need to censor a support tarball. So, the "jumping" will not happen during configuration definition, but only when a problem that requires support intervention, happens. From my PoV this significantly reduces the initial user stress during configuration definition.

Which is important, because when a support tarball is needed there is also going to be "stress", and it is unplanned. 

> 1. Users will provide credentials where they configure the resource - no file jumping during definition -> less chance for configuration errors -> less frustration

if the same resource is used in multiple places, the resource authentication information is duplicated. Also, users might be deploying from a gitops paradigm where the secrets are in a different sensitivity domain. All the customization could be managed in git when the secrets are separated, or the secrets are encrypted in a separate secrets store. 

> 2. Users will not have to manually create files, but instead will just provide their credentials for the resource and we will handle the rest:

That's the case in both cases?

> As I said in the beginning, IMO it comes down to tradeoffs - do we have the user jump through multiple files during configuration (e.g. isolated file thesis), or do we have the user only do the "jumping" if a serious problem that requires outside intervention occurs (e.g. embedded configuration thesis).

Most security guidelines (NIST 800-53, ISO 27001, PCI-DSS) *require* separation of secrets from configuration data.  managing them in place means our solution cannot be compliant with that. 

We will eventually have to offer the option to encrypt secrets for configuration management, and keeping them separate makes that a little bit less awkward imho.

ipetrov117 (commented):
For the thread's context - we have agreed to keep the implementation as it is and integrate over it should user requirements arise. With that in mind I left a couple of comments, let me know what you think!

internal/config/kubernetes.go
```diff
@@ -68,20 +68,21 @@ func (m *Manager) configureKubernetes(
...
-		runtimeHelmCharts, err = m.helm.Configure(conf, manifest)
+		runtimeHelmCharts, runtimeSecrets, err = m.helm.Configure(conf, manifest)
 		if err != nil {
 			return "", "", fmt.Errorf("configuring helm charts: %w", err)
 		}
 	}
 
 	var runtimeManifestsDir string
-	if needsManifestsSetup(conf) {
+	if needsManifestsSetup(conf, runtimeSecrets) {
```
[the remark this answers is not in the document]
ipetrov117: To me this is still a bit off, but I guess this is a personal preference, so I digress. 

Can we however change the parameter naming to something more generic than `runtimeSecrets`, perhaps `additionalManifests`? My main logic here, is that, apart from secrets, there will probably be different types of manifests that we create from code and that would merit a manifests setup.

internal/config/kubernetes_test.go
```diff
@@ -333,5 +344,130 @@ var _ = Describe("Kubernetes", func() {
 			Expect(script).To(BeEmpty())
 			Expect(confScript).ToNot(BeEmpty())
 		})
+
+		It("Succeeds to configure RKE2 with additional resources and auth", func() {
```
ipetrov117: Wouldn't this test fall under the "Succeeds to configure RKE2 with additional resources" test case above? Mainly because Helm charts and repositories can be considered as additional resources.

ipetrov117 (commented):
Great work on this! I think these are my last batch of comments. Once we discuss them, I'll be happy to approve.

ipetrov117 (approved):

dirkmueller (approved):

PR #392, opened 2026-03-18
Adds `release-info` flag to `elemental3`

dharmit:
Example usage:

```sh
$ elemental3 release-info /path/to/manifest.yaml
$ elemental3 release-info /path/to/manifest.yaml --local
$ elemental3 release-info /path/to/manifest.yaml --core
$ elemental3 release-info /path/to/manifest.yaml --product
$ elemental3 release-info /path/to/manifest.yaml -o json
$ elemental3 release-info /path/to/manifest.yaml -o yaml

# All the flags as above, but with an OCI image
$ elemental3 release-info oci-image-containing-release-manifest
...
...
```

This PR doesn't implement the ability to do a diff between two manifests. It'll be added in a subsequent PR.

### DIFF
--- a/cmd/elemental/main.go
+++ b/cmd/elemental/main.go
@@ -37,7 +37,9 @@ func main() {
 		cmd.NewBuildCommand(appName, action.Build),
 		cmd.NewCustomizeCommand(appName, action.Customize),
 		cmd.NewInitCommand(appName, action.Init),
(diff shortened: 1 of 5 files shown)


frelon (approved):
Very nice!

davidcassany (approved):

PR #411, opened 2026-04-09
Fix Kubernetes resource creation retries

atanasdinov:
The previous retry mechanism would false positively return successfully when a single manifest file contains multiple resources and falls under the following conditions:
* at least one fails with "AlreadyExists" error
* all remaining ones fail with a different error

The new approach will instead parse the output line by line to verify that such occurrences are not possible.

### DIFF
--- a/internal/config/templates/k8s_res_deploy.sh.tpl
+++ b/internal/config/templates/k8s_res_deploy.sh.tpl
@@ -12,24 +12,35 @@ retryKubectlCreate() {
   local resource="$1"
   local retries="$2"
   local sleep="$3"
-  local failed=true
+
   for i in $(seq 1 "$retries"); do
     output=$(kubectl_cmd create -f "$resource" 2>&1)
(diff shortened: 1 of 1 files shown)


ldevulder:
Do you thing that this PR could be merged this week or next one? As this is a blocker for the multi-nodes openQA test.

dbw7:
Tested across a variety of VM configurations with our single node deployment example that deploys multiple helm charts and uses the local-path-provisioner manifest that has multiple resources. Everything seemed to deploy properly across each of the VM configurations. I believe we should be fine to rebase and merge once it's marked as ready for review. I can take over and do that if needed

atanasdinov:
@dbw7 Thanks for verifying; there are no conflicts so unless there has been a specific change related to Kubernetes spin-up recently, we should be good to approve and merge as is.

ipetrov117 (approved):

ldevulder (approved):
LGTM

davidcassany (approved):

PR #415, opened 2026-04-15
Remove elemental3ctl sysext image references

davidcassany:
This commits removes the referenes of elemental3ctl being provided as systemd-sysext image.

The generic systemd-sysext documentation is left as is.

### DIFF
--- a/docs/systemd-system-extensions.md
+++ b/docs/systemd-system-extensions.md
@@ -1,11 +1,11 @@
 # System extensions (systemd-sysext) in Elemental
 
-This chapter covers what a system extension (a.k.a. sysext) is, why it's relevant for Elemental, and how Elemental 
+This chapter covers what a system extension (a.k.a. sysext) is, why it's relevant for Elemental, and how Elemental
 uses it to provide extensibility in the image mode OS image.
 
 ## What is a system extension?
 
-[System extension images (or sysext images)](https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html) 
+[System extension images (or sysext images)](https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html)
 can be disk image files or simple folders that get loaded by
 `systemd-sysext.service`. They provide a way to dynamically extend the `/usr`
 and `/opt` directory hierarchies with additional files. When one or more system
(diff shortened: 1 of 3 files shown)


docs/systemd-system-extensions.md
```diff
@@ -147,24 +147,3 @@ Such systemd extensions can be included as an overlay in the Elemental customiza
...
-`elemental3ctl` is a lower-level tool that can do various things like installing an OS (packaged as an OCI image) on a
-target system, upgrading such OS from an OCI image, manage kernel modules on a system, unpack an OCI image, build
-an installation media (generally an ISO file) from an OS image (packaged as OCI image), and more.
-
-`elemental3ctl` is a runtime management tool that helps deploy an OS image on disk, as well as helps manage such an
-installation by performing upgrades, managing kernel modules, perform factory reset, etc. `elemental3` complements
-it by building and configuring an OS image that could have additional artifacts and
-capabilities, making it a platform to build and deploy cloud-native applications.
-
-`elemental3ctl` is provided on an image mode OS as a system extension.
```
dharmit, line 170: IIUC we are not removing the elemental's capability of being able to add sysexts via product manifests. Shouldn't we keep this section as-is in that case? The last statement about `elemental3ctl` being provided as sysext should be removed, of course.

davidcassany: I agree we could document here how systemd extensions can be used in Elemental, however I don't think we can leave this section as is. In fact, if we remove the last statement all the content under `How are systemd extensions used in Elemental?` does not relate to systemd sysexts at all. This disctinction between elemental3 and elemental3ctl is interesting and valid content, but I don't think it belongs here if this is no longer meant to justify elemental3ctl is delievered as sysext. Do you have any idea where we could include this distinction? We can try to find a good spot for it.

Regarding the use of systemd sysext I am bit hesitant to document additional content, its role in Elemental is still a bit blurry.

What you think?

dharmit: You are right about not leaving things as is. The `elemental3`-`elemental3ctl` distinction is something that took me a while to grasp, so that's something I strongly feel about keeping around, as you suggested.

I don't see any single doc talking about both tools. How about moving this info to `index.md` in that case so that new readers see it before almost anything else?

davidcassany: Yes, that might be a good spot, let me try to include it there :+1:

davidcassany: @dharmit added the docs in index.md

dharmit (approved):

frelon (approved):

PR #422, opened 2026-04-24
Support local image extraction from containerd daemon

davidcassany:
This PR adds logic to extract images from containerd local storage. This is meant to only extract images from already unpacked images (uncompressed layer blobs already exist in the filesystem), in fact it essentially mounts the already existing uncompressed tree to a temporary directory to then rsync the contents somewhere else.

Getting into this logic chunk requires two conditions, `--local` flag and `CONTAINERD_SOCK` env variable to define the path of the containerd socket. If no `CONTAINERD_SOCK` variable is defined there is no logic change and it behaves exactly as it used to.

The current PR moves all the logic related to containerd to a new `containerd` package just to make it easier to built upon it in case this is ever needed. Also for the time being I did not code any unit test related to containerd as this tends to be difficult due the client/server nature of it. Once we get this exercised and functional I might try to code some simple test by running a local, temporary and limited containerd server as a go routine just for the test environment.

I did some manual tests of this code following @ipetrov117 suggestion of running a single node deployment including a patched elemental3ctl and then creating a SUC plan which basically runs the OS image in a Pod and sleeps infinitely. Then with a simple `kubectl exec` I got into the pod and validated I could run `elemental3ctl` to upgrade fetching the image from containerd.


go.mod
```diff
@@ -5,7 +5,8 @@ go 1.25.6
 require (
 	dario.cat/mergo v1.0.2
 	github.com/cenkalti/backoff/v4 v4.3.0
-	github.com/containerd/containerd/v2 v2.2.2
+	github.com/containerd/containerd/v2 v2.2.3
+	github.com/containerd/platforms v1.0.0-rc.2
```
atanasdinov: Are we sure we want to settle on a non-official release? If so, RC 4 is already out.

pkg/unpack/oci.go
```diff
@@ -182,3 +218,50 @@ func fetchImage(ctx context.Context, ref name.Reference, platform containerregis
...
+	}
+	return o.onMountedContainerdImage(ctx, callback)
+}
+
+// onMountedContainerdImage mounts as RO the image reference, if found in containerd, and then runs the given callback.
+// Returns the image config digest (imageID) of the image on success or error otherwise. Mountpoints are unmounted and
+// resources freed after executing the method. The given callback gets as input the mountpoint of the image root-tree.
+func (o OCI) onMountedContainerdImage(ctx context.Context, callback func(rootfs string) error) (string, error) {
+	if !o.local {
+		return "", fmt.Errorf("only unpacked images cand be mounted")
```
atanasdinov:
```suggestion
		return "", fmt.Errorf("only unpacked images can be mounted")
```

pkg/unpack/oci.go
```diff
@@ -162,8 +199,7 @@ func (o OCI) Unpack(ctx context.Context, destination string, excludes ...string)
 
 	r := progressbar.NewReader(reader, bar)
 
-	filter := excludesFilter(destination, excludes...)
-	_, err = archive.Apply(ctx, destination, &r, archive.WithFilter(filter))
+	_, err = containerd.Apply(ctx, destination, &r, excludesFilter(destination, excludes...))
```
atanasdinov, line 202: Just to ensure this is intentional; this seems to be executed  when a containerd socket is not specified at all (e.g. line 127). Is this correct behaviour?

atanasdinov (commented):
I've got a few questions but LGTM overall. Happy to approve as soon as those are addressed.

pkg/unpack/oci.go
davidcassany: Yes, this is, in fact, the previous behavior. Since the very first implementation of the image extractor we use the containerd archive apply logic. I moved this call into a wrapper method to encompass all containerd logic and imports into a single package.

go.mod
davidcassany: good catch, gonna double check appropriate version, I guess my env was not fully up to date.

davidcassany: I'd stick with the latest release even if this is RC. This is a really minimalist project to match and define platform names. First RC was a year and half ago this library is rarely touched. Also the [go.mod](https://github.com/containerd/platforms/blob/main/go.mod) that pulls in is really minimalist. I updated to the latest version.

atanasdinov (approved):

ipetrov117 (approved):

pkg/containerd/containerd.go
```diff
@@ -0,0 +1,189 @@
...
+		s:   s,
+	}
+	return cliWrapper, nil
+}
+
+func (w Wrapper) FindUnpackedImage(ctx context.Context, imageRef string) (ImgMeta, error) {
+	var img ImgMeta
+	image, err := w.cli.GetImage(ctx, imageRef)
+	if err != nil {
+		return img, fmt.Errorf("failed to get image from containerd: %w", err)
```
dharmit: At other places in elemental code, we write errors similar to below style/pattern.
```suggestion
		return img, fmt.Errorf("getting image from containerd: %w", err)
```

pkg/containerd/containerd.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	w.s.Logger().Debug("image '%s' found in containerd daemon", imageRef)
```
dharmit:
```suggestion
	w.s.Logger().Debug("image %q found in containerd daemon", imageRef)
```

pkg/containerd/containerd.go
```diff
@@ -0,0 +1,189 @@
...
+
+	introService := w.cli.IntrospectionService()
+	pluginsResp, err := introService.Plugins(ctx, "type==io.containerd.snapshotter.v1")
+	if err != nil {
+		return img, fmt.Errorf("failed to get containerd snapshotter plugins: %w", err)
```
dharmit: Similar suggestion for error: `s/failed to get/getting/`.

pkg/containerd/containerd.go
```diff
@@ -0,0 +1,189 @@
...
+	for _, p := range pluginsResp.Plugins {
+		if ok, _ := image.IsUnpacked(ctx, p.ID); !ok {
+			w.s.Logger().Debug("image not unpacked in %s snapshotter", p.ID)
+			continue
+		}
+		snapshotterName = p.ID
+		break
+	}
+	if snapshotterName == "" {
+		return img, fmt.Errorf("image '%s' is not unpacked", imageRef)
```
dharmit:
```suggestion
		return img, fmt.Errorf("image %q is not unpacked", imageRef)
```

pkg/containerd/containerd.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	cs := w.cli.ContentStore()
+	manifest, err := images.Manifest(ctx, cs, image.Target(), platforms.Default())
+	if err != nil {
+		return img, fmt.Errorf("failed to resolve platform-specific manifest: %w", err)
```
dharmit: s/failed to resolve/resolving

dharmit (commented):
Only comments — suggestions and nits. Nothing to hold the PR from getting merged.

dharmit (approved):

PR #428, opened 2026-04-30
Use compatible Rancher version

dharmit:
Using the current config, Rancher manager installation fails due to RKE2 being too new — 1.35. [Link to support matrix](https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-13-1/).

In my understanding, we could either downgrade the RKE2 version being used for compatibility with Rancher 2.13, or upgrade the Rancher version that works with RKE2 1.35. This PR does the latter.

### DIFF
--- a/examples/elemental/customize/multi-node/suse-product-manifest.yaml
+++ b/examples/elemental/customize/multi-node/suse-product-manifest.yaml
@@ -20,7 +20,7 @@ components:
           crds:
             enabled: true
       - chart: "rancher"
-        version: "2.13.0"
+        version: "2.14.0"
         namespace: "cattle-system"
         repository: "rancher"
         values:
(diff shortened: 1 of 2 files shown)


frelon (approved):
Nice!

examples/elemental/customize/multi-node/suse-product-manifest.yaml
```diff
@@ -20,7 +20,7 @@ components:
           crds:
             enabled: true
       - chart: "rancher"
-        version: "2.13.0"
+        version: "2.14.0"
```
ipetrov117, line 23: We should probably do the same for the [single-node](https://github.com/SUSE/elemental/blob/main/examples/elemental/customize/single-node/suse-product-manifest.yaml#L23-L26) example.

ipetrov117 (approved):
LGTM.

Regarding the Integration Tests [failure](https://github.com/SUSE/elemental/actions/runs/25166815337/job/73775070071?pr=428), I have no idea why, but there is no OBS project for your PR under [devel:UnifiedCore:Tumbleweed:PRs](https://build.opensuse.org/project/subprojects/devel:UnifiedCore:Tumbleweed:PRs), hence the tests fail with 404.

atanasdinov (approved):

PR #429, opened 2026-05-04
Add support for registries file

davidcassany:
This PR adds support for the [registries.yaml](https://docs.rke2.io/install/private_registry) file as part of the `kubernetes/config` folder. The propose implementation of this PR follows a perfectly parallel logic of what it is used for the `server.yaml` and `agent.yaml` files. So that the contents of the file, if present, are deserialized into a ConfigMap and then serialized back again to be embedded as an additional file created as part of the ignition configuration. Feels a bit like an overkill (it could be simply be copied to the overlay directory instead), however I followed this approach to make the rke2 configuration setup handling logic consistent (and arguably easier to maintain) and to also have all the pieces in place assuming we ever want to manipulate the registries file contents as part of the customize process (which I am not seeing at the moment, but probably there is some tweak required for airgap).

### DIFF
--- a/internal/image/kubernetes/cluster.go
+++ b/internal/image/kubernetes/cluster.go
@@ -51,9 +51,16 @@ type Cluster struct {
 	InitServerConfig ConfigMap
 	// AgentConfig contains the agent configurations in multi node clusters.
 	AgentConfig ConfigMap
+	// RegistriesConfig contains the configurations for private or embedded registries
+	RegistriesConfig ConfigMap
 }
 
 func NewCluster(s *sys.System, kube *Kubernetes) (*Cluster, error) {
+	registriesConfig, err := ParseKubernetesConfig(s, kube.Config.RegistriesFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("parsing registries config: %w", err)
+	}
+
 	serverConfig, err := ParseKubernetesConfig(s, kube.Config.ServerFilePath)
 	if err != nil {
 		return nil, fmt.Errorf("parsing server config: %w", err)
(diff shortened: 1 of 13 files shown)


davidcassany:
With this PR it is possible to set server.yaml with:
```yaml
token: totally-not-generated-one
selinux: true
embedded-registry: true
```

and registries.yaml with:

```yaml
mirrors:
  registry.suse.com:
    endpoint:
    - https://registry.suse.com/v2
```

This way on OS upgrades the OS image is only downloaded from the registry once for the first node being updated, then other nodes are pulling it from the embedded registry.

internal/image/kubernetes/cluster.go
```diff
@@ -51,17 +51,27 @@ type Cluster struct {
...
 	// AgentConfig contains the agent configurations in multi node clusters.
 	AgentConfig ConfigMap
+	// RegistriesConfig contains the configurations for private or embedded registries
+	RegistriesConfig ConfigMap
 }
 
 func NewCluster(s *sys.System, kube *Kubernetes) (*Cluster, error) {
+	registriesConfig, err := ParseKubernetesConfig(s, kube.Config.RegistriesFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("parsing agent config: %w", err)
```
atanasdinov:
```suggestion
		return nil, fmt.Errorf("parsing registries config: %w", err)
```

atanasdinov (commented):
LGTM! We only need one minor fix to an error wrapping and to also ensure the linter is happy.

atanasdinov:
Can we also update the customisation directory documentation with the new feature?

davidcassany:
> Can we also update the customisation directory documentation with the new feature?

Sure, docs need to be updated. I also wanted to include the setup in the multinode example.

davidcassany:
Requires a rebase after merging #433 to address lint complaints. Since golangci-lint 2.12 checks for constant are more exhaustive.

ipetrov117 (approved):
LGTM

atanasdinov (approved):

PR #434, opened 2026-05-07
Containerd tests

davidcassany:
This PR adds unit tests for the image extraction from Containerd daemon.

I am not really happy about how messy this turns to be in terms of go modules required just for a couple of tests. Even these dependencies they are not being compiled in the end binary having such a go.mod is a bit frustrating. I wish test dependencies could be explicitly distinguished somehow in go.mod. I looked for alternatives and I did not find any straight forward solution, treating tests as another application with its own go.mod looks like a cumbersome and unpractical approach, specially at development time.


pkg/containerd/containerd_test.go
```diff
@@ -0,0 +1,208 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/suse/elemental/v3/pkg/containerd"
+	"github.com/suse/elemental/v3/pkg/log"
+	"github.com/suse/elemental/v3/pkg/sys"
+	sysmock "github.com/suse/elemental/v3/pkg/sys/mock"
+	"github.com/suse/elemental/v3/pkg/sys/vfs"
+
+	"github.com/containerd/containerd/defaults"
```
atanasdinov: Seems like this one should be using `v2` instead which should get rid of some of the go.mod entries.

davidcassany: Good catch, thanks :+1: indeed, this should have not been here, however this does not improve the indirect dependencies required... 😞

ipetrov117 (approved):

dharmit (approved):
Learnt more from this PR than "reviewing" it. :)

atanasdinov (approved):

frelon (approved):

PR #436, opened 2026-05-07
Drop name field from release.yaml

atanasdinov:
Simplifies the release.yaml schema dropping the optional "name" field which doesn't provide too much of a value.

### DIFF
--- a/examples/elemental/customize/linux-only/release.yaml
+++ b/examples/elemental/customize/linux-only/release.yaml
@@ -1,2 +1,1 @@
-name: suse-product
 manifestURI: oci://registry.suse.com/beta/uc/release-manifest:0.6_rke2_1.35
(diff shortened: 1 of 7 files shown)


examples/elemental/customize/linux-only/release.yaml
```diff
@@ -1,2 +1 @@
-name: suse-product
 manifestURI: oci://registry.suse.com/beta/uc/release-manifest:0.6_rke2_1.35
```
atanasdinov, line 1: This kind of makes me think that we need to reconsider the config dir structure. Having a single file to specify only the manifest URI for Linux deployments feels too complex from UX perspective.

ipetrov117 (approved):

davidcassany (approved):

PR #442, opened 2026-05-12
Fix overlay dir with split mode

dbw7:
Currently when you run split mode with no configuration that creates an overlay dir (e.g. kubernetes), the image build fails because it expects an overlay dir that doesn't exist. The new change now makes sure not to set `d.OverlayTree = overlaySource` if the overlay dir doesn't exist, which allows the image build to succeed.

### DIFF
--- a/internal/customize/customize.go
+++ b/internal/customize/customize.go
@@ -220,12 +220,10 @@ func parseDeployment(
 		d.BootConfig.KernelCmdline = fips.AppendCommandLine(d.BootConfig.KernelCmdline)
 	}
 
-	overlaysURI := fmt.Sprintf("%s://%s", deployment.Dir, output.OverlaysDir())
-	overlaySource, err := deployment.NewSrcFromURI(overlaysURI)
-	if err != nil {
-		return nil, fmt.Errorf("parsing overlay source URI %q: %w", overlaysURI, err)
+	overlaysDir := output.OverlaysDir()
+	if exists, _ := vfs.Exists(fs, overlaysDir); exists {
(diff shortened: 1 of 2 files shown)


ipetrov117 (approved):

internal/customize/customize.go
```diff
@@ -220,12 +220,15 @@ func parseDeployment(
...
-	if err != nil {
-		return nil, fmt.Errorf("parsing overlay source URI %q: %w", overlaysURI, err)
+	overlaysDir := output.OverlaysDir()
+	if exists, _ := vfs.Exists(fs, overlaysDir); exists {
+		overlaysURI := fmt.Sprintf("%s://%s", deployment.Dir, overlaysDir)
+		overlaySource, err := deployment.NewSrcFromURI(overlaysURI)
+		if err != nil {
+			return nil, fmt.Errorf("parsing overlay source URI %q: %w", overlaysURI, err)
+		}
+		d.OverlayTree = overlaySource
```
davidcassany: These lines could be further simplified with something like:

```suggestion
		d.OverlayTree = deployment.NewDirSrc(overlaysDir)
```

davidcassany (commented):
Looks good, but these specific lines could be slightly simplified.

internal/customize/customize.go
dbw7: Thank you! I tested this and it seemed to work on the 4 configurations I tried, much cleaner.

davidcassany (approved):

ipetrov117 (approved):

frelon (approved):

PR #443, opened 2026-05-18
HTTP Helm Auth Fix

dbw7:
This PR fixes Helm authentication for HTTP repositories. The current logic only properly works for OCI repositories, this should fix that.

### DIFF
--- a/internal/config/helm.go
+++ b/internal/config/helm.go
@@ -46,7 +46,7 @@ type helmChart interface {
 	GetName() string
 	GetInlineValues() map[string]any
 	GetRepositoryName() string
-	ToCRD(values []byte, repository string, hasAuth bool) *helm.CRD
+	ToCRD(values []byte, repository string, hasAuth, skipTLSVerify bool) *helm.CRD
 }
 
 type Helm struct {
@@ -152,8 +152,10 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
 	}
 
 	for _, chart := range charts {
-		needsAuth := authMap[chart.Chart] != nil
-		if err = h.appendHelmChart(chart, repositories, valueFiles, &crds, needsAuth); err != nil {
+		a := authMap[chart.Chart]
+		needsAuth := a != nil
+		skipTLSVerify := needsAuth && a.InsecureSkipTLSVerify
+		if err = h.appendHelmChart(chart, repositories, valueFiles, &crds, needsAuth, skipTLSVerify); err != nil {
 			return nil, nil, fmt.Errorf("collecting helm charts: %w", err)
 		}
 	}
@@ -163,8 +165,10 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
 		valueFiles = conf.Kubernetes.Helm.ValueFiles()
 
 		for _, chart := range conf.Kubernetes.Helm.Charts {
-			needsAuth := authMap[chart.Name] != nil
-			if err = h.appendHelmChart(chart, repositories, valueFiles, &crds, needsAuth); err != nil {
+			a := authMap[chart.Name]
+			needsAuth := a != nil
+			skipTLSVerify := needsAuth && a.InsecureSkipTLSVerify
+			if err = h.appendHelmChart(chart, repositories, valueFiles, &crds, needsAuth, skipTLSVerify); err != nil {
 				return nil, nil, fmt.Errorf("collecting user helm charts: %w", err)
 			}
 		}
@@ -186,12 +190,15 @@ func createAuthMap(charts []*api.HelmChart, repositories map[string]string, conf
 			if !ok || rc.Credentials == nil {
 				continue
 			}
-			extractedHost, err := extractHost(repositories[c.Repository])
+
+			repoURL := repositories[c.Repository]
+			extractedHost, err := extractHost(repoURL)
 			if err != nil {
 				return nil, fmt.Errorf("extracting host: %w", err)
 			}
 			authMap[c.Chart] = &auth.HelmAuth{
-				URL: extractedHost,
+				RawURL: repoURL,
+				URL:    extractedHost,
 				Credentials: auth.Credentials{
 					Username: rc.Credentials.Username,
 					Password: rc.Credentials.Password,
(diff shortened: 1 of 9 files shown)


frelon (approved):

davidcassany (approved):

internal/config/helm.go
```diff
@@ -152,8 +152,13 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
 	}
 
 	for _, chart := range charts {
-		needsAuth := authMap[chart.Chart] != nil
-		if err = h.appendHelmChart(chart, repositories, valueFiles, &crds, needsAuth); err != nil {
+		a := authMap[chart.Chart]
+		needsAuth := a != nil
+		skipTLSVerify := false
```
ipetrov117: nit:
```suggestion
		skipTLSVerify := needsAuth && a.InsecureSkipTLSVerify
```

internal/config/helm.go
```diff
@@ -163,8 +168,13 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
 		valueFiles = conf.Kubernetes.Helm.ValueFiles()
 
 		for _, chart := range conf.Kubernetes.Helm.Charts {
-			needsAuth := authMap[chart.Name] != nil
-			if err = h.appendHelmChart(chart, repositories, valueFiles, &crds, needsAuth); err != nil {
+			a := authMap[chart.Name]
+			needsAuth := a != nil
+			skipTLSVerify := false
```
ipetrov117: nit:

```suggestion
			skipTLSVerify := needsAuth && a.InsecureSkipTLSVerify
```

internal/image/kubernetes/kubernetes.go
```diff
@@ -93,14 +93,15 @@ func (c *HelmChart) GetRepositoryName() string {
...
 }
 
 type HelmRepository struct {
-	Name        string            `yaml:"name" validate:"required"`
-	URL         string            `yaml:"url" validate:"required,url"`
-	Credentials *auth.Credentials `yaml:"credentials,omitempty"`
+	Name                  string            `yaml:"name" validate:"required"`
+	URL                   string            `yaml:"url" validate:"required,url"`
+	Credentials           *auth.Credentials `yaml:"credentials,omitempty"`
+	InsecureSkipTLSVerify bool              `yaml:"insecureSkipTLSVerify,omitempty"`
```
ipetrov117, line 104: Would it make sense to also include this in the [manifest/api/shared.go](https://github.com/SUSE/elemental/blob/main/pkg/manifest/api/shared.go#L128) as this essentially relates more to the URL than the actual credentials, IIRC. That way we would have this configuration in the product and core release manifest.

Although I am not sure whether we would get a use-case where a consumer would define a release manifest pointing to a repository that requires a skip in the TLS verification, it could be useful to keep the release manifest and `cluster.yaml` Helm repository configurations in sync. Either way, could be done in a follow-up PR.

pkg/helm/helm.go
```diff
@@ -70,18 +72,21 @@ type SecretMetadata struct {
...
+	DockerConfigJSON *string `yaml:".dockerconfigjson,omitempty"`
+	Username         *string `yaml:"username,omitempty"`
+	Password         *string `yaml:"password,omitempty"`
 }
 
-func NewCRD(namespace, chart, version, valuesContent string, repository string, auth bool) *CRD {
+func NewCRD(namespace, chart, version, valuesContent string, repository string, auth, skipTLSVerify bool) *CRD {
 	name := chart
-
+	isOCI := false
```
ipetrov117: nit:

```suggestion
	isOCI := strings.HasPrefix(repository, "oci://")
```

internal/config/helm.go
```diff
@@ -332,33 +351,48 @@ func enabledHelmCharts(rm *resolver.ResolvedManifest, enabled []release.HelmChar
...
 			Name:      fmt.Sprintf("%s-auth", name),
 			Namespace: "kube-system",
 		},
-		Type: "kubernetes.io/dockerconfigjson",
-		Data: helm.SecretData{
-			DockerConfigJSON: encoded,
-		},
 	}
+
+	if creds.IsOCI {
```
ipetrov117: Instead of having the whole `HelmAuth.IsOCI` logic for this single check, can we just do a `strings.HasPrefix(creds.URL, "oci://")` here and drop the whole `HelmAuth.IsOCI` logic from the struct? 

This would also align with what we do under [helm/helm.go](https://github.com/dbw7/elemental/blob/helm-auth-fix/pkg/helm/helm.go#L83).

ipetrov117 (approved):
Thanks for doing this! I left out a couple of comments, if they make sense we could handle them in a follow-up PR. Either way, good job.

internal/image/kubernetes/kubernetes.go
dbw7: I agree, in addition to this gap, there's a few other gaps that all need to be addressed in a follow PR.

dbw7:
Thank you for the reviews, I've made some updates and changes. There are some gaps remaining that need to be addressed in a new PR but I don't want them to block this PR for right now if you guys are alright with that. The biggest gap is that the internal logic currently ties `insecureSkipTLSVerify` to repositories that have authentication, this means that if it's set to `true` for a non-authenticated repository, it will be ignored, so it only works for authenticated repos at the moment.

The next large gap is using `insecureSkipTLSVerify` is a band-aid fix, ideally, we need to support allowing a user to provide a certificate that can be used in the Helm controller for additional verification instead of just skipping the cert check, but I wasn't sure how much will need to change to support this, so I opted for the fastest unblocking fix.

pkg/helm/helm.go
```diff
@@ -70,12 +72,14 @@ type SecretMetadata struct {
...
+	Username         *string `yaml:"username,omitempty"`
+	Password         *string `yaml:"password,omitempty"`
 }
 
-func NewCRD(namespace, chart, version, valuesContent string, repository string, auth bool) *CRD {
+func NewCRD(namespace, chart, version, valuesContent string, repository string, auth, skipTLSVerify bool) *CRD {
 	name := chart
-
+	isOCI := strings.HasPrefix(repository, "oci://")
 	if strings.HasPrefix(repository, "oci://") {
```
ipetrov117: nit: can be fixed in the follow-up PR.

```suggestion
	if isOCI {
```

ipetrov117 (approved):

ipetrov117 (approved):

PR #450, opened 2026-05-21
Do not force relabel persistent volumes

davidcassany:
This PR performs system relabel only for the immutable parts of the system.

This change is to prevent changing labels for files which are potentially in use at the time they are relabeled. This is to protect upgrades from runtime race conditions. Persistent volumes should be relabeled on first boot after upgrading and not by elemental3ctl upgrade command.

In addition the PR also renames the SE Linux methods to SystemRelabel and ChrootedSystemRelabel. The reason for that is that those methods are assuming a full system root tree relabel, so they are not supposed to be used to relabel any given path, that's why I added the `System`, if we ever need to relabel a specific path only we can always extend the package API.

I also removed the possibility of adding additional bind mounts for the chrooted call (e.g. to mount in the tree some additional partition or volume) as we are only using this method during the upgrade process and in that process the root-tree already has all the mounts set in place and so we were not using this feature at all. This feature was the responsible of having some additional checks and convoluted logic as part of the chrooted relabel, since this was unused I removed it.

### DIFF
--- a/pkg/installer/media.go
+++ b/pkg/installer/media.go
@@ -567,7 +567,7 @@ func (i Media) prepareOSRoot(sourceOS *deployment.ImageSource, rootDir string) e
 		return fmt.Errorf("OS source write failed: %w", err)
 	}
 
-	err = selinux.Relabel(i.ctx, i.s, rootDir)
+	err = selinux.SystemRelabel(i.ctx, i.s, rootDir)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


ipetrov117 (dismissed):
Nice!

frelon (approved):

ipetrov117 (approved):

PR #452, opened 2026-05-22
Bump Go to 1.26

atanasdinov:


ipetrov117 (approved):

davidcassany (approved):
LGTM, however for the integration tests to pass it requires manual intervention in OBS to update the specfile.

This is something I'd like to bring up in a design meeting, adding the specfile in this repository, this way we can have it all in sync with a single PR without having to manually sync OBS and IBS. Far from being an urgent topic, but something I'd like to discuss.

dirkmueller:
> LGTM, however for the integration tests to pass it requires manual intervention in OBS to update the specfile.

.. which we can do before this PR. the issue here was that the services VM was not new enough, so no amount of updating of the spec file would have helped here.

PR #453, opened 2026-05-22
Trigger .autorelabel if there were paths excluded

davidcassany:
This PR generates the /etc/selinux/.autorelabel file if there were some paths excluded during the relaballing at upgrade time. Persistent shared paths are excluded to prevent runtime race conditions. With this change after every upgrade the system gets relabelled at boot (feature provided by `selinux-autorelabel` on SUSE systems). Since each snapshot has its own snapshotted /etc the relabelling only happens when the new updated snapshot boots for the first time.


davidcassany:
Draft as this is based on another PR (#455) , gonna rebase it from main as soon as #455 gets merged and set it as ready to review.

frelon (approved):

ipetrov117 (approved):

PR #455, opened 2026-05-22
SE Linux: distinguish snapshotted vs shared RW paths

davidcassany:
When relabelling a new OS snapshot this commits sets a force relabelling for all immutable parts and a non forced relabelling for RW and snapshotted paths (e.g. /etc). These are also a new snapshot hence there is no risk of interfering with the running system. Shared persistent volumes (e.g. /var) are excluded from this relabelling as they could be already in use by some other process in the running system.

### DIFF
--- a/pkg/selinux/selinux.go
+++ b/pkg/selinux/selinux.go
@@ -20,7 +20,9 @@ package selinux
 import (
 	"container/ring"
 	"context"
+	"fmt"
 	"path/filepath"
+	"slices"
 
 	"github.com/suse/elemental/v3/pkg/chroot"
 	"github.com/suse/elemental/v3/pkg/sys"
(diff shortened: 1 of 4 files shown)


atanasdinov (approved):

frelon (approved):

pkg/selinux/selinux.go
```diff
@@ -31,20 +33,23 @@ const (
 	SelinuxTargetedContextFile = selinuxTargetedPath + "/contexts/files/file_contexts"
 
 	selinuxTargetedPath = "/etc/selinux/targeted"
+	selinuxAutoRelabel  = "/etc/selinux/.autorelabel"
```
ipetrov117: nit: probably a leftover from #453.

ipetrov117 (approved):

pkg/selinux/selinux.go
davidcassany: Yes removing this line.

PR #457, opened 2026-05-25
Add lifecycle metadata to product/solution and core release manifests

jtomasek:
Adds additional metadata to solution and core release manifests
    
- Adds `lifecycle` section to solution and core release manifests
- In case when lifecycle is defined, at least `lifecycle.availabilityDate` must be set
- Bumps release manifest `schema` version to `v1`, defaults to `v0` if not provided
- Removed `metadata.creationDate` from release manifests, replaced by `lifecycle.availabilityDate`
- Updated tests
- Updated example manifests
- Updated docs

### DIFF
--- a/docs/release-manifest.md
+++ b/docs/release-manifest.md
@@ -7,6 +7,10 @@ Ultimately, there are two types of release manifests:
 * [Solution Release Manifest](#solution-release-manifest)
 * [Core Platform Release Manifest](#core-platform-release-manifest)
 
+## Schema versions
+
+Manifests may declare a schema version with a top-level `schema` field. Both `v0` and `v1` are accepted; when the field is omitted, the manifest is treated as `v0`. The `lifecycle` section is optional; when supplied, `availabilityDate` is required and all other lifecycle fields remain optional.
+
 ## Solution Release Manifest
 
 > **NOTE:** Elemental is in active development and the Solution manifest API may change over time.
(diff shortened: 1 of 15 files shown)


jtomasek:
There are a few things I'd like to get feedback on:

We probably want to make the lifecycle section optional and document it's purpose.

The design document also defines `lifecycle.stage` property. Updating it however requires re-releasing of the specific release manifest image which IIUC is not desirable. We could update the `lifecycle.stage` property in the codebase without updating the image however I don't think this is something we want to do.

I am quite inclined towards using the Kubernetes Object Model for the release manifest format as it clearly defines 
- Type Metadata: (`apiVersion` and `kind`) to define the manifest versioning
- Object Metadata: A `metadata` block for standardized labels and annotations which could be easily utilised for lifecycle properties
- Specification: A `spec` block containing the `corePlatform` and `components` definition.

This approach would also make the product/solution definition close to the EIB based Edge release manifest definition, making the transition more straightforward.

docs/release-manifest.md
```diff
@@ -20,10 +29,16 @@ Enables consumers to extend a specific `Core Platform` release with additional c
...
+schema: v1
 metadata:
   name: "SUSE Product"
   version: "4.2.0"
   creationDate: "2025-07-10"
+lifecycle:
+  availabilityDate: "2025-07-10"
+  fullSupportEndDate: "2026-07-10"
+  maintenanceSupportEndDate: "2027-07-10"
+  eolDate: "2028-07-10"
```
fdegir: Should we add another property to state the type of the release, e.g., `lts`?

jtomasek: Done

PR #471, opened 2026-05-28
Upgrade docs

davidcassany:


docs/filesystem.md
```diff
@@ -121,6 +121,45 @@ This means:
 - Local configuration changes persist across updates via the merge process
 - Rolling back the OS also rolls back `/etc` to match that OS version
 
+#### 3 Way Merge on Snapshotted Directories
+
+The three way merge concept is common in distributed version control systems such as git. It essentially describes
+the process of merging two branches. The same concept can easily be applied to directory trees. In Elemental the
+essential focus is around the contents of /etc in a running system.
```
dharmit: We're putting /etc in codeblocks across this doc otherwise.
```suggestion
essential focus is around the contents of `/etc` in a running system.
```

docs/filesystem.md
```diff
@@ -121,6 +121,45 @@ This means:
...
+
+The merge consists on applying two independent deltas over the same origin. There are the stock default contents of
+the current image in which there are some customization applied on top (e.g. users configurations, ssh keys, etc.), this
+is the customizations delta. On the other side there is the delta between the stock default contents of the current image
+versus the defaults ofi the new image, the defaults delta.
```
dharmit: Typo
```suggestion
versus the defaults of the new image, the defaults delta.
```

docs/filesystem.md
```diff
@@ -121,6 +121,45 @@ This means:
...
+| exists            | modified         | not modified     | new default file |
+| exists            | modified         | modified         | customized file  |
+| exists            | modified         | deleted          | file deleted     |
+| exists            | not modified     | modified         | customized file  |
+| exists            | not modified     | deleted          | file deleted     |
+| exists            | deleted          | not modified     | file deleted     |
+| exists            | deleted          | modified         | customized file  |
+| missing           | default creation | custom creation  | custom creation  |
+| missing           | default creation | missing          | default creation |
+| missing           | missing          | custom creation  | custom creation  |
```
dharmit: What does "creation" means in this context?

docs/filesystem.md
```diff
@@ -121,6 +121,45 @@ This means:
...
+
+#### Merge combinations
+
+Customized files win over new default files. The following matrix aims to illustrate a variety of combinations
+and the expected outcome in each case.
+
+| Original defaults | New defaults     | Customized state | Merge result     |
+| ----------------- | ---------------- | ---------------- | ---------------- |
+| exists            | modified         | not modified     | new default file |
+| exists            | modified         | modified         | customized file  |
```
dharmit, line 151: What if the new defaults add some configuration directive to the file that has been modified in customized state? However, the customized state doesn't have that directive configured.

Will the merge result have the configuration that was added via new defaults?

docs/index.md
```diff
@@ -34,5 +34,7 @@ capabilities, making it a platform to build and deploy cloud-native applications
 * [Building a Linux Image](building-linux-image.md) - for users and/or consumers interested in building Linux images.
 * [Image Customization](image-customization.md) - for users and/or consumers interested in customizing images that are based on a specific release.
 * [Release Manifest Guide](release-manifest.md) - for consumers interested in creating a release manifest for their solution.
+* [Configuration Directory Guide](configuration-directory.md) - for users and/or consumers interested on checking configration options.
```
dharmit:
```suggestion
* [Configuration Directory Guide](configuration-directory.md) - for users and/or consumers interested in checking configration options.
```

docs/filesystem.md
```diff
@@ -121,14 +121,48 @@ This means:
...
+the current image in which there are some customization applied on top (e.g. users configurations, ssh keys, etc.), this
+is the customizations delta. On the other side there is the delta between the stock default contents of the current image
+versus the defaults ofi the new image, the defaults delta.
+
+The merging process applies both deltas over the original content with preference for the customizations delta.
+This is a two step process, first applies the defaults delta and then applies the customizations delta. Deltas are
+defined at file granularity, meaning the entire file is copied when the delta is applied. This way customized files are
+always kept during upgrades, having preference over the defaults delta in case of overlaps. Customized files always win.
+
+> [!NOTE]
```
dharmit: Unfortunately, these admonitions don't get recognized by Jekyll. So this will look perfect on GitHub's interface, but look borked on opensource.suse.com/elemental.

dharmit (commented):
Some nits and couple of questions.

docs/filesystem.md
davidcassany: It won't customized state has precedence. The merge concept of two directory trees is at file granularity, you pick files from one or the other. There is no merge at _patch_ level.

docs/filesystem.md
davidcassany: Oh, then I'll just use a quote with NOTE title, thanks!

docs/filesystem.md
dharmit: Makes sense. Thanks for explaining.

docs/filesystem.md
dharmit: I understand that we are talking about a scneario where a file didn't exist in the original image shipped by us. But I don't understand what "default creation" and "custom creation" mean.

davidcassany: Yes, I am aware this matrix it might be not obvious enough to read, I'd be happy to switch to a better or more obvious nomenclature. I was in doubt in many of these terms so suggestions are more than welcome.

So, in this line, `missing` means the file under analysis that was not present in the original image. Then `custom creation` for the `Customized state` means the user added it or created it via some sort of configuration. `default creation` for the `New defaults` means the new image added this file as part of its contents. We could use `addition` instead of creation, hence `default addition` or  `custom addition` or any other concept we can think of.

dharmit: > So, in this line, `missing` means the file under analysis that was not present in the original image. Then `custom creation` for the `Customized state` means the user added it or created it via some sort of configuration. `default creation` for the `New defaults` means the new image added this file as part of its contents.

Thanks. This makes it clear for me. 👍🏽 


> We could use `addition` instead of creation, hence `default addition` or `custom addition` or any other concept we can think of.

I'd suggest `newly created` or `newly added` in place of `default addition` and `user created` in place of `custom addition`. Also, put the scenario with both `default creation` and `custom creation` last in this list.

dharmit (approved):

ipetrov117 (approved):

PR #473, opened 2026-05-29
Create CHANGELOG.md

dbw7:


CHANGELOG.md
```diff
@@ -0,0 +1,215 @@
+# Changelog
+
+## v3.1.0-alpha.20260528
```
davidcassany, line 3: Note this will have to be adapted when backporting to branch release-3.0.

@dirkmueller I just now realized that tag `v3.1.0-alpha...` is also part of the release-3.0 as this points to the commit from where the branch was actually created, so whatever we tag in release-3.0 will have parent tag of `v3.1.0-alpha` I don't think this will cause as any issue now as we are not using tags in any way AFAIK. However this looks wrong and confusing by looking at the history of the release-3.0 branch.

davidcassany (approved):
Nice! looks good to me :+1:

I wonder if we could figure out a way to automate a changes generation draft PR

frelon (approved):

CHANGELOG.md
dbw7: Thank you for the feedback, ok so I'm down to make any necessary changes, I'll assume this is ok to merge here and we can make any tweaks in a small PR for whatever we need for the branch and tag

PR #477, opened 2026-06-01
elemental3ctl: accept new iso filename as well

dirkmueller:


fdegir (approved):

ipetrov117 (approved):

davidcassany (approved):

atanasdinov (approved):

frelon (approved):

internal/cli/action/customize.go
```diff
@@ -143,7 +143,7 @@ func setupConfigManager(s *sys.System, configDir string, output config.Output, l
 }
 
 func setupFileExtractor(ctx context.Context, s *sys.System, outDir config.Output, local bool) (extr *extractor.OCIFileExtractor, err error) {
-	const isoSearchGlob = "/iso/uc-base-kernel-default-iso*.iso"
+	const isoSearchGlob = "/iso/*default-iso*.iso"
```
ldevulder, line 146: It could have been a good idea to add support for RT kernel as well, no?

atanasdinov: Yes, but we can add this later. I wouldn't want to preemptively try to think of the exact name before we settle on how to structure it first.

PR #478, opened 2026-06-01
elemental3ctl: accept new iso filename as well

dirkmueller:


ipetrov117 (approved):

davidcassany (approved):

atanasdinov (approved):

internal/cli/action/customize.go
```diff
@@ -143,7 +143,7 @@ func setupConfigManager(s *sys.System, configDir string, output config.Output, l
 }
 
 func setupFileExtractor(ctx context.Context, s *sys.System, outDir config.Output, local bool) (extr *extractor.OCIFileExtractor, err error) {
-	const isoSearchGlob = "/iso/uc-base-kernel-default-iso*.iso"
+	const isoSearchGlob = "/iso/*default-iso*.iso"
```
frelon, line 146: I would consider matching on /iso/*.iso, does this work for -rt kernels? 🤔

dirkmueller: it doesn't work for rt kernels, obviously (but right now we do not even have that). 

I'd be in favor of making the glob even more liberal like you suggested.

dirkmueller:
the test is passing locally when run on an openSUSE tumbleweed host:

```


Ran 6 of 6 Specs in 81.606 seconds
SUCCESS! -- 6 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

```

PR #481, opened 2026-06-02
Fix Kubernetes Node Handling

### LINKED ISSUE #437: server.yaml file ignored on single-node cluster
For an image generated with `customize` command, I have this `kubernetes/config/server.yaml` file defined:
```yaml
token: qa-test-token
debug: true
selinux: true
```

On a multi-node cluster this file is copied on the installed OS in `/etc/rancher/rke2/config.yaml` but on a single-node cluster it seems that this file is simply ignored. Could be an issue if we want to define some option with it.

dbw7:
Closes #437

### DIFF
--- a/internal/config/kubernetes.go
+++ b/internal/config/kubernetes.go
@@ -182,7 +182,7 @@ func writeK8sConfigDeployScript(fs vfs.FS, output Output, k kubernetes.Kubernete
 		err      error
 	)
 
-	if len(k.Nodes) > 0 {
+	if len(k.Nodes) > 1 {
 		initNode, err = kubernetes.FindInitNode(k.Nodes)
 		if err != nil {
 			return "", fmt.Errorf("finding init node: %w", err)
--- a/internal/config/kubernetes_test.go
+++ b/internal/config/kubernetes_test.go
@@ -345,6 +345,51 @@ var _ = Describe("Kubernetes", func() {
 			Expect(confScript).ToNot(BeEmpty())
 		})
 
+		It("Uses server config for a single explicitly configured server node", func() {
+			conf := kubernetes.Kubernetes{
+				Nodes: kubernetes.Nodes{
+					{Hostname: "node01", Type: kubernetes.NodeTypeServer},
+				},
+			}
+
+			confScript, err := writeK8sConfigDeployScript(
+				fs,
+				output,
+				conf,
+				"/opt/k8s/install",
+				"/opt/k8s/install/install.sh",
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			b, err := fs.ReadFile(filepath.Join(output.OverlaysDir(), confScript))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(b)).To(ContainSubstring(`CONFIGFILE="/var/lib/elemental/kubernetes/${NODETYPE}.yaml"`))
+			Expect(string(b)).ToNot(ContainSubstring("init.yaml"))
+		})
+
+		It("Uses init config only for multi-node clusters", func() {
+			conf := kubernetes.Kubernetes{
+				Nodes: kubernetes.Nodes{
+					{Hostname: "server01", Type: kubernetes.NodeTypeServer},
+					{Hostname: "agent01", Type: kubernetes.NodeTypeAgent},
+				},
+			}
+
+			confScript, err := writeK8sConfigDeployScript(
+				fs,
+				output,
+				conf,
+				"/opt/k8s/install",
+				"/opt/k8s/install/install.sh",
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			b, err := fs.ReadFile(filepath.Join(output.OverlaysDir(), confScript))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(b)).To(ContainSubstring(`if [[ "${HOSTNAME}" = "server01" ]]; then`))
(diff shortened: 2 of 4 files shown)


rdoxenham:
Pulled this patch in and deployed a single node with the following `cluster.yaml` and `server.yaml`:

```
% cat kubernetes/cluster.yaml
nodes:
- hostname: uc-test.rancher.local
  type: server
network:
    apiVIP: 192.168.122.100
    apiHost: 192.168.122.100.sslip.io

% cat kubernetes/config/server.yaml
token: totally-not-generated-one
selinux: true
cni: cilium
embedded-registry: true
```

And it worked just fine...

```
# cat /etc/rancher/rke2/config.yaml
cni: cilium
embedded-registry: true
selinux: true
tls-san:
    - 192.168.122.100
    - 192.168.122.100.sslip.io
token: totally-not-generated-one
```

I still need to try a multi-node just to ensure that still works, but it certainly fixes the issue recorded in #437. Thanks!!

frelon (approved):
Nice work!

dharmit:
> Pulled this patch in and deployed a single node with the following `cluster.yaml` and `server.yaml`:
> 
> ```
> % cat kubernetes/cluster.yaml
> nodes:
> - hostname: uc-test.rancher.local
>   type: server
> network:
>     apiVIP: 192.168.122.100
>     apiHost: 192.168.122.100.sslip.io
> 
> % cat kubernetes/config/server.yaml
> token: totally-not-generated-one
> selinux: true
> cni: cilium
> embedded-registry: true
> ```
> 
> And it worked just fine...

I am unable to grasp what's wrong  because I see this even with current `main` branch. However, the single-node VM created with this configuration still has `single-node-example` as its hostname which is something I feel is wrong because:
1. We set `hostname: uc-test.rancher.local` in `cluster.yaml`.
2. When spinning up the VM using virt-install, I used the following command:
    ```sh
    sudo virt-install --name node01 \
    --ram 16384 \
    --vcpus 6 \
    --osinfo detect=on,name=sle-unknown \
    --graphics none \
    --console pty,target_type=serial \
    --network network=default,model=virtio,mac=FE:C4:05:42:8B:01 \
    --virt-type kvm \
    --import \
    --boot uefi,loader=/usr/share/qemu/ovmf-x86_64-ms-4m-code.bin,nvram.template=/usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin \
    --disk path="/var/lib/libvirt/images/main.raw",format=raw
    ```
    So I was expecting the resulting VM's hostname to either be `uc-test.rancher.local` or `node01`, but it's `single-node-example`. 😕

rdoxenham:
> 
> I am unable to grasp what's wrong because I see this even with current `main` branch. However, the single-node VM created with this configuration still has `single-node-example` as its hostname which is something I feel is wrong because:
> 
> 1. We set `hostname: uc-test.rancher.local` in `cluster.yaml`.

It will only match on this hostname if you have a `uc-test.rancher.local` entry in `network/`.

The default examples have a specific entry for `single-node-example` (note that this filename is expected to translate into a desired **hostname**). When `nmc` loads it will try and match based on mac address to understand which node is which, hence why it sets your hostname to `single-node-example`:

```
rdo@tiw network % pwd
(snip) elemental/examples/elemental/customize/single-node

rdo@tiw network % grep mac-address single-node-example.yaml
    mac-address: FE:C4:05:42:8B:01
```

You will see that this mac address matches your `virt-install` call:

> 2. When spinning up the VM using virt-install, I used the following command:
>    ```shell
>    sudo virt-install --name node01 \
>    --ram 16384 \
>    --vcpus 6 \
>    --osinfo detect=on,name=sle-unknown \
>    --graphics none \
>    --console pty,target_type=serial \
>    --network network=default,model=virtio,mac=FE:C4:05:42:8B:01 \
>    --virt-type kvm \
>    --import \
>    --boot uefi,loader=/usr/share/qemu/ovmf-x86_64-ms-4m-code.bin,nvram.template=/usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin \
>    --disk path="/var/lib/libvirt/images/main.raw",format=raw
>    ```
>      
>    So I was expecting the resulting VM's hostname to either be `uc-test.rancher.local` or `node01`, but it's `single-node-example`. 😕

So, to fix this, rename your `single-node-example.yaml` file in `network/` to `uc-test.rancher.local`.

You have, however, stumbled on an issue: should we block the install if the available nodes *don't* match the desired node names in `cluster.yaml`, @atanasdinov @dirkmueller ?

atanasdinov:
We used to fail in [EIB](https://github.com/suse-edge/edge-image-builder/blob/5ad4c9b8647edf0e8e092212f8a42ff437cc634a/pkg/combustion/templates/k3s-multi-node-installer.sh.tpl#L22), however, the implementation for [Elemental](https://github.com/SUSE/elemental/blob/be4a555ace67e1e30967b10f0331d902bc6fd496/internal/config/templates/k8s_conf_deploy.sh.tpl#L16) will just default to server.

The latter is more forgiving, but I guess it's something to discuss once again if it proves unexpected.

ldevulder (approved):
LGTM. Manually tested the fix on my lab, works as expected!

dharmit:
Thanks @rdoxenham for elaborate responses. 🙏🏽 

> You have, however, stumbled on an issue: should we block the install if the available nodes _don't_ match the desired node names in `cluster.yaml`

I spent unreasonable time yesterday and today trying to understand what's going on here. At least something (a bug that caused me to not face another bug) useful came out of it. 😂

atanasdinov (approved):

PR #508, opened 2026-06-24
Detect conflicts during merge conflict

atanasdinov:
During an upgrade, /etc (and any snapshotted RW volume) is reconciled with a 3-way merge: the customizations delta (old stock -> user-modified) is applied on top of the new image's defaults. When a file is changed by both the user and the new image, the user's version wins, but until now that happened silently.

This change detects those overlaps and warns the user, e.g.:
```
Merge conflicts detected for /etc (2 file(s)), user version kept:
    /modifiedFile — user: modified, OS: modified
    /deletedFile — user: deleted, OS: modified
```

Previously users had no visibility into cases where an upgrade wanted to change a file they had also customised. The merge result is correct (customisations are intentionally preferred), but silently discarding the new image's version of a config file is surprising. This surfaces it so the user can reconcile manually if needed.

**The merge outcome is unchanged.** This is detection and reporting only; no automated resolution, no change to what ends up on disk:
  - The customisations delta is still applied exactly as before (user version always wins).
  - The process of volume configurations and stock-snapshot handling, xattr filtering, and rsync invocations are all unchanged.
  - Conflict detection is a read-only per-file comparison (old stock vs new image) done in the same single pass that already parses the snapper status file and applies changes; no extra tree walk, no second parse as per David's initial review.

One implementation nuance in `applyCustomChanges`: user deletions are now buffered and applied after the status scan instead of inline. This is required so the per-file OS-delta check sees a pristine new tree. It is equivalent to the previous behaviour for the on-disk result; deleted paths and synced paths are disjoint, and both still complete before rsync runs.
  
For each file the user changed (from snapper status), we report whether the new image also changed that path (added/modified/deleted, by type → size → content, with a size cap for large blobs). If both sides touched it, it's recorded as a conflict and reported via a warning after the volume is merged.

### DIFF
--- a/pkg/merge/merge.go
+++ b/pkg/merge/merge.go
@@ -1,0 +1,191 @@
+/*
+Copyright © 2025-2026 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package merge
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	iofs "io/fs"
+	"os"
+	"strings"
+
+	"github.com/suse/elemental/v3/pkg/sys/vfs"
+)
+
+// ChangeType represents the kind of change made to a file relative to a
+// common ancestor. The zero value is the empty string, deliberately not a
+// valid change, so an uninitialised ChangeType reads as obviously missing.
+type ChangeType string
+
+const (
+	ChangeTypeAdded    ChangeType = "added"
+	ChangeTypeModified ChangeType = "modified"
+	ChangeTypeDeleted  ChangeType = "deleted"
+)
+
+// MaxContentCompareSize bounds how large a regular file may be before
+// FileChange stops comparing its content byte-for-byte. Files above this
+// size whose stat metadata matches are reported as modified rather than
+// read in full; this trades a small false-positive risk for bounded I/O
+// on user-supplied blobs.
+const MaxContentCompareSize = 16 * 1024 * 1024
+
+// Conflict represents a file the user and the OS both changed relative to
+// the common ancestor.
+type Conflict struct {
+	Path       string
+	UserChange ChangeType
+	OSChange   ChangeType
+}
+
+func (c Conflict) String() string {
+	return fmt.Sprintf("  %s — user: %s, OS: %s", c.Path, c.UserChange, c.OSChange)
+}
+
+// FormatConflictSummary returns a human-readable summary of detected
+// conflicts; returns "" if there are none.
+func FormatConflictSummary(volumePath string, conflicts []Conflict) string {
+	if len(conflicts) == 0 {
+		return ""
+	}
+
+	var b strings.Builder
+
+	_, _ = fmt.Fprintf(&b, "Merge conflicts detected for %s (%d file(s)), user version kept:\n",
+		volumePath, len(conflicts))
+	for _, c := range conflicts {
+		_, _ = fmt.Fprintf(&b, "%s\n", c)
+	}
+
+	return b.String()
+}
+
+// FileChange reports how a single file changed between oldPath and newPath.
+// Returns empty string if the file is unchanged (or absent from both sides).
+//
+// Regular files are compared by size and then content (up to
+// MaxContentCompareSize; larger matched-size files are conservatively
+// flagged as modified without being read). Symlinks compared by target.
+// Type changes (file->dir, file->symlink, ...) reported as modified.
+// Directory-only additions/deletions are ignored (contents at those paths
+// are what's meaningful, and snapper reports them individually).
+func FileChange(fs vfs.FS, oldPath, newPath string) (ChangeType, error) {
+	oldInfo, oldErr := fs.Lstat(oldPath)
+	if oldErr != nil && !os.IsNotExist(oldErr) {
+		return "", oldErr
+	}
+	newInfo, newErr := fs.Lstat(newPath)
+	if newErr != nil && !os.IsNotExist(newErr) {
+		return "", newErr
+	}
+
+	oldExists := oldErr == nil
+	newExists := newErr == nil
+
+	switch {
+	case !oldExists && !newExists:
+		return "", nil
+	case !oldExists:
+		if newInfo.IsDir() {
+			return "", nil
+		}
+		return ChangeTypeAdded, nil
+	case !newExists:
+		if oldInfo.IsDir() {
+			return "", nil
+		}
+		return ChangeTypeDeleted, nil
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/transaction/snapper_upgrade_helper.go
davidcassany, line 329: Wouldn't it be possible to check for conflict on file per file basis here in this loop instead of walking the whole tree? So here we just check the user modified tree if it is also modified by the new image. So we only diff the files which are potentially conflicting and we do not need to parse again snapper diff status, load it in memory, iterate over it again and walk the whole tree of the snapshot.
So we basically create  the list of files to merge and the list of conflicting files together at the same time.

pkg/transaction/transaction.go
```diff
@@ -52,6 +52,7 @@ type Merge struct {
 	Old      string // old unmodified tree
 	New      string // new base tree where modifications should be applied
 	Modified string // modified tree on top of the old tree
+	NewStock string // new stock snapshot path (for conflict detection)
```
davidcassany: I don't think this is needed. This has the same content as `New` at merge time.

davidcassany (commented):
I appreciate this is in draft review, hence this is probably not ready for review. If you don't mind I have a couple of comments regarding the implementation.

I am wondering if instead of gathering a full diff for both cases wouldn't it be simpler to diff one and then from a first list of modified files then use it to diff them on the other side of the 3 way merge. After all we want to detect conflicts so any file not included in the first diff is known not to be a conflict candidate.

pkg/transaction/transaction_test.go
```diff
@@ -204,7 +206,8 @@ func snapperContextMock() {
 	runner = sysmock.NewRunner()
 	tfs, cleanup, err = sysmock.TestFS(nil)
 	Expect(err).NotTo(HaveOccurred())
-	logger := log.New(log.WithDiscardAll())
+	logBuf = &bytes.Buffer{}
+	logger := log.New(log.WithBuffer(logBuf))
```
davidcassany, line 210: This buffer is initialized but not used right?. IIUC the diff of this file is most likely a leftover, this does not apply any effective change.

pkg/merge/merge.go
```diff
@@ -0,0 +1,191 @@
...
+	ChangeTypeModified ChangeType = "modified"
+	ChangeTypeDeleted  ChangeType = "deleted"
+)
+
+// MaxContentCompareSize bounds how large a regular file may be before
+// FileChange stops comparing its content byte-for-byte. Files above this
+// size whose stat metadata matches are reported as modified rather than
+// read in full; this trades a small false-positive risk for bounded I/O
+// on user-supplied blobs.
+const MaxContentCompareSize = 16 * 1024 * 1024
```
davidcassany, line 47: Does it mean any file bigger than 16MiB which is modified in both branches (upgraded content and modified content) is always considered a conflict? Just to make sure I am understanding it.

I wonder if 16MiB is low value. In any case I'd not expect many big and potentially conflicting files under /etc, as it would mean we ship them at first place. Have you checked if we ship and big file under /etc?

davidcassany (approved):
LGTM, just a couple of minor comments

pkg/transaction/transaction_test.go
atanasdinov: It is used in the assertions in `snapper_upgrade_helper_test.go` (line 270+).

davidcassany: Oh yes, sorry, I forgot this is shared code with several tests. I just looked in was not used within the same file 😅

pkg/merge/merge.go
atanasdinov: Yes, your understanding is correct, with one caveat.

`FileChange` is only evaluated for files snapper already flagged as user-modified, and it compares the old stock vs the new image to decide the OS side. For a file that exists on both sides and is larger than 16 MB, it reports "modified" without reading the content:
  - if the sizes differ -> the file genuinely changed.
  - if the sizes are equal -> the file is assumed modified rather than reading the whole blob.

So for such a file, "user modified it" + "assumed OS-modified" = conflict, even in the (rare) case where the new image's copy is byte-identical. That's the trade-off we introduce here: bounded I/O vs. a possible false-positive warning (resolution is unaffected as user's version is kept either way).

On whether 16 MB is low value, this is what we are observing for the largest files (ignoring /etc/.snapshots):

```
  ┌──────┬─────────────────────────────────────────────────────────────────┐
  │ Size │                              File                               │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 13M  │ /etc/udev/hwdb.bin                                              │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 5.3M │ /etc/selinux/targeted/contexts/files/file_contexts.bin          │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 3.8M │ /etc/selinux/targeted/policy/policy.34                          │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 436K │ /etc/selinux/targeted/contexts/files/file_contexts              │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 296K │ /etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 116K │ /etc/mime.types                                                 │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 88K  │ /etc/vmware-tools/vgauth/schemas/XMLSchema.xsd                  │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 32K  │ /etc/nftables/osf/pf.os                                         │
  ├──────┼─────────────────────────────────────────────────────────────────┤
  │ 20K  │ /etc/vmware-tools/tools.conf.example, /etc/grub.d/20_linux_xen  │
  └──────┴─────────────────────────────────────────────────────────────────┘
```

In general, I'd expect files over 16MB to be binaries and deeming them as "large" enough not to consume memory is probably a safer bet. Thoughts?

ipetrov117 (approved):
Overall approach looks good to me, approved 🚀 . I think we should consider how this will be handled in the LCM use-case as well (not in this PR obviously). 

Right now LCM runs `elemental3 upgrade ..` through a SUC workload. If a conflict is detected, it will be logged, but no-one will be able to act upon it, because merge conflicts will not break the upgrade. As the upgrade will be marked as successful, the SUC workload will reboot the OS and only after this reboot will the user be aware that something is wrong due to a potentially breaking conflict. 

I wonder whether it would make sense to make failing on merge conflicts configurable. That way LCM could block an upgrade on merge conflicts, allowing for the user to inspect the conflicts and only after assessing their severity, instruct LCM to continue with the upgrade. This would ensure that we catch user configuration conflict errors before we actually apply the new version, which will reduce the chance for needlessly recovering an OS after an upgrade.

I recognise that this **will** have some caveats, but I think it is worth considering, as blindly accepting merge conflicts on an automated upgrade process could prove to be problematic. That said I can be missing, misunderstanding, or overcomplicating things. Let me know what you think.

atanasdinov:
Yeah, fully agreed. This is the initial step towards robust handling for conflict detection and resolution. For now, we only get to print the differences. In the future, we can define policies on how this is being handled, and whether the user wants to abort the upgrade process with an error if conflicts are detected. TBD.

davidcassany:
Merging as the a rebased new PR with a clean new OBS project creation passed all tests smoothly https://github.com/SUSE/elemental/pull/554 The test failure in this PR is due to some leftover in OBS, the PR has been around for quite a while and the associated OBS project might have some inconsistency.

PR #522, opened 2026-07-08
Stackable initrd

davidcassany:
This PR extends the bootloader interface and the deployment struct to support additional initrds to be stacked at install time.

With this change `elemental3ctl` is capable to concatenate initrds with a setup such as:
```yaml
bootloader:
  name: grub
  kernelCmdline: "console=ttyS0"
  initrdExtensions:
  - /path/to/initrd/extension.cpio
```

Extensions a pre-appended to the initrd found in the OS image, this prevents extensions to modify any pre-existing file, it can only append.

Note for the reviewers, easier to follow changes by looking commits individually, note most of the line changes are caused by a simple bootloader API refactor to make it more readable and extendable.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -230,6 +230,10 @@ type Disk struct {
 type BootConfig struct {
 	Bootloader    string `yaml:"name"`
 	KernelCmdline string `yaml:"kernelCmdline"`
+
+	// InitrdExtensions represents a list of CPIO files which are added in the
+	// bootloader initrd call in addition to the stock initrd included within the OS
+	InitrdExtensions []string `yaml:"initrdExtensions,omitempty"`
 }
 
 type FirmwareConfig struct {
@@ -926,6 +930,11 @@ func (d *Deployment) WriteDeploymentFile(s *sys.System, root string) error {
 	dep.OverlayTree = nil
 	dep.CfgScript = ""
 	dep.Installer = LiveInstaller{}
+
+	// omit initrd extensions as this is a runtime information which might not be consistent on reboots
+	if dep.BootConfig != nil {
+		dep.BootConfig.InitrdExtensions = nil
+	}
 
 	data, err := yaml.Marshal(dep)
 	if err != nil {
--- a/pkg/deployment/deployment_test.go
+++ b/pkg/deployment/deployment_test.go
@@ -171,11 +171,18 @@ var _ = Describe("Deployment", Label("deployment"), func() {
 			d := deployment.DefaultDeployment()
 			d.Disks[0].Device = "/dev/device"
 			d.SourceOS = deployment.NewDirSrc("/some/image")
+			d.BootConfig = &deployment.BootConfig{
+				Bootloader:       "grub",
+				KernelCmdline:    "kernel parameters",
+				InitrdExtensions: []string{"/path/to/initrd/extension"},
+			}
 			Expect(d.WriteDeploymentFile(s, "/some/dir")).To(Succeed())
 			rD, err := deployment.Parse(s, "/some/dir")
 			Expect(err).NotTo(HaveOccurred())
 			Expect(len(rD.Disks)).To(Equal(1))
 			Expect(rD.Disks[0].Device).To(BeEmpty())
+			Expect(rD.BootConfig).NotTo(BeNil())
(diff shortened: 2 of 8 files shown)


pkg/deployment/deployment_test.go
```diff
@@ -171,11 +171,18 @@ var _ = Describe("Deployment", Label("deployment"), func() {
...
+				KernelCmdline:    "kernel parameters",
+				InitrdExtensions: []string{"/path/to/initrd/extension"},
+			}
 			Expect(d.WriteDeploymentFile(s, "/some/dir")).To(Succeed())
 			rD, err := deployment.Parse(s, "/some/dir")
 			Expect(err).NotTo(HaveOccurred())
 			Expect(len(rD.Disks)).To(Equal(1))
 			Expect(rD.Disks[0].Device).To(BeEmpty())
+			Expect(rD.BootConfig).NotTo(BeNil())
+			Expect(rD.BootConfig.InitrdExtensions).To(BeEmpty())
```
atanasdinov, line 185: On the system the list of extensions should not be visible?

atanasdinov (approved):

pkg/deployment/deployment_test.go
davidcassany: Indeed, this points to some of the weaknesses of using the same schema for installation time, upgrade time and to track current status. Since these extensions are not part of the system they are no longer available at runtime. Storing them somewhere in the installed system is also something tricky, how and when should they be used?
In current implementation what is considered to be an install time  data (e.g. disk device names, initrd extensions) is not recorded in `/etc/elemental/deployment.yaml` so they are not loaded in upgrades. With the current approach previous initrd extensions setup would never be merged during upgrades, it would only consider the ones explicitly defined with the install|upgrade call.

davidcassany:
This PR concatenates initrds instead of executing multiple ones in the bootloader setup. I went for a concatenation because it is less intrusive code wise and it is probably easier to integrate with FDE. In any case it could be changed to calling separate files easily, I have no strong opinion. The bootloader interface would remain untouched.

pkg/deployment/deployment.go
```diff
@@ -230,6 +230,10 @@ type Disk struct {
 type BootConfig struct {
 	Bootloader    string `yaml:"name"`
 	KernelCmdline string `yaml:"kernelCmdline"`
+
+	// InitrdExtensions represents a list of CPIO files which are added in the
+	// bootloader initrd call in addition to the stock initrd included within the OS
+	InitrdExtensions []string `yaml:"initrdExtensions,omitempty"`
```
ipetrov117, line 236: I wonder whether we should validate for duplicate filenames here, or have some predictable naming when we store the CPIO files under the `Install` directory.

My main concern is that we could have a bootloader that points to two different initrd extensions that are on a different path but of the same name, something like:
```yaml
bootloader:
  ...
  initrdExtensions:
  - /path/to/initrd/extension.cpio
  - /path/to/initrd2/extension.cpio
```

When [copying](https://github.com/davidcassany/elemental/blob/stackable_initrd/pkg/installer/media.go#L644) these CPIO files to the `Install` directory, we will have the second entry overwrite the first entry, due to them having the same name. Leading to the first entry not being used later when we concatenate the files.

Furthermore, when we [append](https://github.com/davidcassany/elemental/blob/stackable_initrd/pkg/installer/media.go#L650) the extensions we will have the same entry defined twice, something like:

```bash
[]InitrdExtensions{"<root>/Install/extension.cpio", "<root>/Install/extension.cpio"}
```

Which, while not fatal, would lead to the same entry being concatenated twice at install time.

I may be misreading the code, so please let me know what you think. Also, I am happy with handling this in a follow-up PR, should we decide that this is actually an issue.

ipetrov117 (commented):
Looks clean. I just have one question after which I will be happy to approve.

pkg/deployment/deployment.go
davidcassany: Good point. Even if with the current UX this not possible, I think, mostly because this logic copying to `Install` directory is only executed as part of `customize` and in customize we are not exposing this configuration options. In any case I agree this should be improved. I think the good solution is to simply add a predictable prefix to the copied files. I am adding this improvement here in this PR :wink:

ipetrov117 (approved):

PR #523, opened 2026-07-08
CPIO files utility

davidcassany:
This PR extends the runner interface to support piped input and outputs and it also creates a small internal utility to create and extract CPIO files.

### DIFF
--- a/pkg/sys/mock/runner.go
+++ b/pkg/sys/mock/runner.go
@@ -83,7 +83,13 @@ func (r *Runner) RunContextWithPipe(
 
 	callErrChan := make(chan error, 1)
 	go func() {
-		_, e := io.Copy(stdout, pr)
+		var e error
+		if stdout != nil {
+			_, e = io.Copy(stdout, pr)
+		}
+		if r.ReturnError != nil {
+			e = r.ReturnError
+		}
 		callErrChan <- e
 	}()
 
--- a/pkg/sys/mock/runner.go
+++ b/pkg/sys/mock/runner.go
@@ -21,6 +21,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/suse/elemental/v3/pkg/log"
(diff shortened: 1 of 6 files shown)


davidcassany:
This PR introduces a call to `cpio` utility in the scope of `elemental3 customize` so that this implies adding `cpio` rpm as a requirement of `elemental3` package to ensure it is present in the elemental3 image.

I did not consider using a golang library (like [this one](https://github.com/cavaliergopher/cpio)) as I considered it was not worth adding an extra go dependency while having a fully supported `cpio` tool in almost any linux OS out there.

pkg/sys/mock/runner.go
```diff
@@ -69,6 +70,37 @@ func (r *Runner) RunContext(_ context.Context, command string, args ...string) (
 	return r.Run(command, args...)
 }
 
+func (r *Runner) RunContextWithPipe(
```
atanasdinov, line 73: ReturnError/SideEffect values are being ignored. Do we want to add them as a proper unit test failure for mocked cpio errors?

pkg/sys/mock/runner.go
```diff
@@ -69,6 +70,37 @@ func (r *Runner) RunContext(_ context.Context, command string, args ...string) (
...
+	var err error
+
+	r.cmds = append(r.cmds, append([]string{command}, args...))
+	r.envs = append(r.envs, append([]string{command}, envs...))
+
+	pr, pw := io.Pipe()
+
+	callErrChan := make(chan error, 1)
+	go func() {
+		_, e := io.Copy(stdout, pr)
```
atanasdinov: I believe this can panic if `stdout` is not set. Is it an issue?

pkg/sys/runner/runner_test.go
```diff
@@ -126,4 +131,40 @@ var _ = Describe("Runner", Label("runner"), func() {
...
+			return err
+		}
+		workDir, err = tfs.RawPath(workDir)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Reads first line from the standard input and the second line from a fine in the working directory
+		Expect(r.RunContextWithPipe(context.Background(), callback, buff, nil, workDir, nil, "cat", "-", "secondLine")).To(Succeed())
+		Expect(buff.String()).To(ContainSubstring("First line\nSecond line\n"))
+	})
+	It("runs a command getting input from a pipe in a specific working directory", func() {
```
atanasdinov: The test name is duplicated from above.

pkg/sys/runner/runner.go
```diff
@@ -145,6 +146,56 @@ func (r run) RunContextParseOutput(ctx context.Context, stdoutH, stderrH func(st
...
+	}
+
+	if err = cmd.Start(); err != nil {
+		_ = stdinPipe.Close()
+		return fmt.Errorf("%q command reported an error on start: %w", command, err)
+	}
+
+	if err = stdinPipeFn(stdinPipe); err != nil {
+		_ = stdinPipe.Close()
+		_ = cmd.Wait()
```
atanasdinov: This call should contain the real error that we need to report. When failing, `err = stdinPipeFn(stdinPipe)` is likely to contain `io: read/write on closed pipe` because the subprocess would exit before consuming all of stdin.

One approach could be to keep both errors in the propagated error, e.g. `return fmt.Errorf("callback returned an error: %w, pipe closed with: %w", cErr, err)`.

Have you ran into such issue?

atanasdinov (commented):
Great submission! I got a few comments and questions, but LGTM overall.

pkg/sys/runner/runner.go
davidcassany: Ah, good point, yes there could be a meaningful error provided by the command itself. A command failure breaks the pipe and then I just get the broken pipe error, however the relevant issue is on the command returned error. I am adding it, thanks

pkg/sys/mock/runner.go
davidcassany: I'd entirely ignore SideEffect, as it is not obvious to me how to keep a similar meaningful functionality in this case, but ReturnError can be improved. I am on it.

atanasdinov (approved):

ipetrov117 (approved):

PR #528, opened 2026-07-13
Stock ignition configuration

davidcassany:
This PR adds the `--mode merge` and `--mode splitmerge` in `elemental3 customize`. If mode `merge` or  `splitmerge` is set Elemental will embed the Ignition configuration as an initrd extension. With this option the ignition configuration will be seen as base configuration from Ignition's PoV and allow to merge on top additional user configuration. Moreover, the configuration partition, if any at all, will be labeled with `CATALYST` instead of `ignition` if no `<mnt>/ignition/config.ign` file is present. `merge` mode keeps the rest of the configuration embedded in a configuration partition. `splitmerge` generates non ignition configuration in a configuration directory separate from the image.

This is handy in scenarios where the user configuration is provided by the platform instead of being embedded within the image itself. Consider cases such as deploying in cloud providers.

### DIFF
--- a/internal/cli/cmd/customize.go
+++ b/internal/cli/cmd/customize.go
@@ -44,7 +44,7 @@ func NewCustomizeCommand(appName string, action func(context.Context, *cli.Comma
 		Usage:     "Customize an image based on a release",
 		UsageText: fmt.Sprintf("%s customize", appName),
 		Before: func(ctx context.Context, _ *cli.Command) (context.Context, error) {
-			modes := []string{"", "embedded", "split"}
+			modes := []string{"", "embedded", "split", "merge", "splitmerge"}
 			if !slices.Contains(modes, CustomizeArgs.Mode) {
 				return ctx, cli.Exit("Error: Unsupported --mode option.", 1)
 			}
@@ -74,8 +74,10 @@ func NewCustomizeCommand(appName string, action func(context.Context, *cli.Comma
 			},
 			&cli.StringFlag{
 				Name: "mode",
-				Usage: "Customization mode, 'embedded' (config partition within image) or " +
-					"'split' (configuration directory separate from the image)",
+				Usage: "Customization mode, 'embedded' (config partition within image), " +
+					"'split' (configuration directory separate from the image), " +
+					"'merge' (ignition config in initrd and the rest in a config partition) or " +
+					"'splitmerge' (ignition config in initrd and the rest in a separate directory)",
 				Destination: &CustomizeArgs.Mode,
 				Value:       "embedded",
 			},
--- a/internal/cli/cmd/customize.go
+++ b/internal/cli/cmd/customize.go
@@ -44,7 +44,7 @@ func NewCustomizeCommand(appName string, action func(context.Context, *cli.Comma
 		Usage:     "Customize an image based on a release",
 		UsageText: fmt.Sprintf("%s customize", appName),
 		Before: func(ctx context.Context, _ *cli.Command) (context.Context, error) {
-			modes := []string{"", "embedded", "split", "merge", "splitmerge"}
+			modes := []string{"", "embedded", "split"}
 			if !slices.Contains(modes, CustomizeArgs.Mode) {
 				return ctx, cli.Exit("Error: Unsupported --mode option.", 1)
 			}
@@ -74,10 +74,8 @@ func NewCustomizeCommand(appName string, action func(context.Context, *cli.Comma
 			},
 			&cli.StringFlag{
 				Name: "mode",
-				Usage: "Customization mode, 'embedded' (config partition within image), " +
-					"'split' (configuration directory separate from the image), " +
-					"'merge' (ignition config in initrd and the rest in a config partition) or " +
-					"'splitmerge' (ignition config in initrd and the rest in a separate directory)",
+				Usage: "Customization mode, 'embedded' (config partition within image) or " +
+					"'split' (configuration directory separate from the image)",
 				Destination: &CustomizeArgs.Mode,
 				Value:       "embedded",
 			},
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN ARCH=$(uname -m); \
     zypper --non-interactive install --no-recommends xfsprogs \
         util-linux-systemd \
         e2fsprogs \
+        cpio \
         udev \
         rsync \
         grub2 \
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -21,6 +21,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/suse/elemental/v3/internal/config"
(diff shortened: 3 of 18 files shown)


davidcassany:
This PR is based on top of #522 and #523, because of that it is marked as draft. The diff will be way easier to read and review once #522 and #523 are part of the main branch.

internal/cli/cmd/customize.go
```diff
@@ -90,6 +91,11 @@ func NewCustomizeCommand(appName string, action func(context.Context, *cli.Comma
 				Usage:       localDesc,
 				Destination: &CustomizeArgs.Local,
 			},
+			&cli.BoolFlag{
+				Name:        "base-config",
```
ipetrov117: I am probably missing something, but from a UX perspective wouldn't it make more sense for this to be moved under the `--mode` flag, as this is essentially a customization mode?

My concern is that having both `--mode` and `--base-config` would cause confusion in users, because from their perspective both flags specify a way to configure how ignition and additional configurations are shipped. This may lead to users specifying incorrect flag relationships. For example, specifying `--mode split` with `--base-config` will produce an extracted configuration directory, while also having the ignition config  embedded as base along with the additional config under the `CATALYST` partition.

Moving the `--base-config` flag as an option to `--mode` would remove this potential problem while keeping the UX fairly clean.

We could have something like `--mode ignition-base`, or we could extend the existing `embedded` option to be more verbose. For example have `--mode embedded=partition` for configurations (ignition + additional) embedded in the image under the `ignition` partition and `--mode embedded=base` for configurations where ignition is embedded as the base configuration while any additional configurations are defined under the `CATALYST` partition.

You probably had a good reason for not merging those flags, so please let me know what you think.

ipetrov117 (commented):
Thanks for the work on this @davidcassany! I left one suggestions that I would be interested to hear your thoughts on.

Apart from that, I hit an error when trying to test the changes. Below I tried to explain what I found, but the TL;DR is that it seems to be related to how we handle already defined initrd extensions during the installer's boot environment.

Error hit when doing `./elemental3ctl --debug install` at install time:

```bash
INFO[0033] Installing kernel/initrd
INFO[0033] Reading OS Release
DEBU[0033] Concatenating extensions [/run/initramfs/live/Install/0-0-initrdExt.cpio] and initrd "/tmp/elemental_system158623787/@/.snapshots/1/snapshot/usr/lib/modules/6.12.0-160000.35-default/initrd"
ERRO[0033] Closing transaction due to a failure: installing bootloader: installing kernel+initrd: copying initrd '/tmp/elemental_system158623787/@/.snapshots/1/snapshot/usr/lib/modules/6.12.0-160000.35-default/initrd': stat /run/initramfs/live/Install/0-0-initrdExt.cpio: no such file or directory
[  243.306303][ T2088] EXT4-fs (sda3): unmounting filesystem e41514e2-03df-4660-adb3-cc64998c0971.
```

IIUC, this is happening because at install time we setup the recovery partition and run the [addInstallationAssets](https://github.com/davidcassany/elemental/blob/stock_ignition_configuration/pkg/installer/media.go#L641-L655) boot config logic, which again attempts to copy the files from what is defined under `initrdExtensions` to `/run/initramfs/live/Install/<index>-<entry-as-seen-in-initrdExtensions>`. However, because the `Deployment` object at install time is populated from the `/run/initramfs/live/Install/install.yaml` file, which is already defined like so:

```yaml
...
bootloader:
    ...
    initrdExtensions:
        - /run/initramfs/live/Install/0-initrdExt.cpio
```

the function adds an extra `0` index to the `0-initrdExt.cpio` extension, which produces an invalid name that makes the underlying [ConcatFiles](https://github.com/davidcassany/elemental/blob/stock_ignition_configuration/pkg/sys/vfs/vfs.go#L573) function to fail with the mentioned error. 

Please let me know whether I am doing something wrong, both when customising the image and operating in the installer boot environment, I am using an `elemental3ctl` binary built with this PR's changes.

davidcassany:
> Apart from that, I hit an error when trying to test the changes. Below I tried to explain what I found, but the TL;DR is that it seems to be related to how we handle already defined initrd extensions during the installer's boot environment.

Oh damn, yes I under estimated the change to rename files with a prefix to prevent overlaps, this logic gets called twice and causes the glitch, my bad. I am looking at it.

internal/cli/cmd/customize.go
davidcassany: Yes, this is up for discussion. My first intention was to actually allow both at the same time. `--mode split` moves the contents of `ignition` partition (catalyst and ignition files) to a separate directory without creating the partition as part of the image. `--base-config` sets all generated ignition configuration to be embedded as part of the initrd, no `ignition` partition. If we do a `--mode split` and `--base-config` the result should be a separate directory with catalyst only configuration and the ignition configuration being embedded as part of the initrd. No `ignition` partition anywhere.

>  For example, specifying --mode split with --base-config will produce an extracted configuration directory, while also having the ignition config embedded as base along with the additional config under the CATALYST partition.

Let me double check they are not interacting badly. My intention was essentially to allow split mode having ignition configuration already embedded in the initrd. So this gives room to easily add on top additional ignition configuration at run time to an image which does not include any configuration partition.

I am happy to extend `--mode` to allow all four combinations, even if we do not map all of them I believe this should be only a UX constrain.

ipetrov117: Ah, I missed the fact that `config.ign` is not part of the split directory when we combine both options. 

> > For example, specifying --mode split with --base-config will produce an extracted configuration directory, while also having the ignition config embedded as base along with the additional config under the CATALYST partition.
>
> Let me double check they are not interacting badly. My intention was essentially to allow split mode having ignition configuration already embedded in the initrd. So this gives room to easily add on top additional ignition configuration at run time to an image which does not include any configuration partition.

What I was thinking about when I wrote this was a scenario where we add some "default" configuration as base, but then the underlying provisioning mechanism (say CAPI)  expects only its configurations to be applied. I was wondering whether there was a case where some of those configurations would clash and cause confusion. For example  a user defining a `kubernetes` directory and then specifying `--mode split` + `--base-config` which, IIUC, would lead to us embedding our kubernetes systemd service logic as the base config, but the user then wants to deploy this image on CAPI. Admittedly this is a user error, but we allowed for this use-case in the first place.

That was the main reason  I suggested merging this in the `--mode` flag. As then everything will be explicitly defined under one place and  we would not open ourselves up to such corner cases. That said, I may be over engineering this, or missing something.. 

> I am happy to extend --mode to allow all four combinations, even if we do not map all of them I believe this should be only a UX constrain.

These combination would be:
1. Ignition + added configurations in an `ignition` partition (current `embedded` mode)
2. Ignition + added configurations outside of the image (current `split` mode)
3. Ignition set as base configuration + added configurations in `CATALYST` partition (new mode)
4. Ignition set as base configuration + added configurations in a directory (current `split` + `base` config)

^ Am I correct in these definitions? 

Taking a step back, I think we should briefly discuss this on one of our design sessions, as taking the right approach to UX is always tricky and subjective. 

Apologies for derailing your PR with this, if you feel like this can be discussed as a follow-up, I am happy to do that as well.

davidcassany: Yes these are the four combinations, 3 and 4 are the new ones this PR aims to introduce. I am fine discussing it in a design discussion. Meanwhile I'll update the PR to extend `--mode` to the four types, I believe this will make the discussion easier. IMHO we should code thinking for all 4 and then consider if it makes sense exposing all them to the user, to me these are orthogonal options.

davidcassany:
@ipetrov117 the regression seen on ISO installation is fixed. Moreover I also extended the `mode` flag with two additional modes `merge` and  `splitmerge`. This way the current behavior remains unchanged and it adds some more flexibility on how to define ignition configuration. We can follow up on a UX discussion what and how we want to expose base configuration to the user.

ipetrov117 (approved):
@davidcassany thanks, this looks great!

davidcassany:
As discussed I changed the implementation to always set ignition configuration as stock base configuration. So no further modes have been applied and two existing ones changed their behavior to always embed the ignition configuration into the initrd. This allows to extend and/or overwrite ignition configuration at run time by merging the user configuration on top.

ipetrov117 (approved):

atanasdinov (approved):

PR #534, opened 2026-07-20
New release schema version

davidcassany:
This PR is easier to review commit by commit. First commit refactors a bit the versioning approach. It basically moves the schema definition into a private sub-package while keeping the public definition as an alias to the latest version. Then when adding a new version on a follow up commit the mapping needs to be updated to the latest version and the migration logic should be incorporated.

The new version is only redefining the structs that have been modified any untouched structure is simply aliased to the new version, which in turn will be aliased to the latest one.

The changes in the schema removes the version in Metadata for solution and core manifests and it also introduces the `Elemental` field for core manifests. 

```yaml
schema: v1
metadata:
  name: "foobar"
  creationDate: "20260720"
  elemental:
    version: "3.1"
    image: "registry.com/foo/bar/elemental:3.1-4.23"
components:
```

If defined `elemental` metadata requires both, version and image.


pkg/manifest/api/solution/manifest.go
```diff
@@ -18,53 +18,62 @@ limitations under the License.
...
+type CorePlatform = solutionv1.CorePlatform
+type Components = solutionv1.Components
 
-type CorePlatform struct {
-	Image string `yaml:"image" validate:"required"`
-}
+func Parse(data []byte) (*ReleaseManifest, error) {
+	version, err := api.LoadSchemaVersion(data)
+	if err != nil {
+		return nil, fmt.Errorf("parsing 'core' release manifest: %w", err)
```
ipetrov117:
```suggestion
		return nil, fmt.Errorf("parsing 'solution' release manifest: %w", err)
```

pkg/manifest/api/core/manifest.go
```diff
@@ -18,64 +18,65 @@ limitations under the License.
...
+	default:
+		return nil, fmt.Errorf("unknown release manifest version %q", version)
 	}
+}
 
-	if err := api.NewValidator(api.WithYAMLFieldNames()).Struct(rm); err != nil {
-		var validationErrors validator.ValidationErrors
-		if errors.As(err, &validationErrors) {
-			err = api.FormatErrors(validationErrors)
+func migrateV0(old *corev0.ReleaseManifest) (*corev1.ReleaseManifest, error) {
```
ipetrov117: This function never returns an error, do we want to drop it as a return value?

pkg/manifest/api/solution/manifest.go
```diff
@@ -18,53 +18,62 @@ limitations under the License.
...
+		return parseV1(data)
+	default:
+		return nil, fmt.Errorf("unknown release manifest version %q", version)
+	}
 }
 
-func Parse(data []byte) (*ReleaseManifest, error) {
-	if _, err := api.LoadSchemaVersion(data); err != nil {
-		return nil, fmt.Errorf("parsing 'solution' release manifest: %w", err)
+func migrateV0(old *solutionv0.ReleaseManifest) (*solutionv1.ReleaseManifest, error) {
```
ipetrov117: This function never returns an error, do we want to drop it as a return value?

ipetrov117 (approved):
LGTM, minor nits.

pkg/manifest/api/core/manifest_test.go
```diff
@@ -147,7 +153,7 @@ components:
 		rm, err := core.Parse(data)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(rm).ToNot(BeNil())
-		Expect(rm.Schema).To(BeEquivalentTo("v0"))
+		Expect(rm.Schema).To(BeEquivalentTo("v1"))
```
ipetrov117, line 153: nit: do we also want to validate that the new fields introduced from `v1` are parsed as well?

pkg/manifest/api/solution/manifest.go
davidcassany: Yes, I thought about keeping it as probably the migration logic evolves and becomes complex, just to make explicit this could eventually return an error at some point. But can remove it and eventually re-add it if needed.

ipetrov117: Ah, makes sense. I am fine keeping it explicit 👍

ipetrov117 (approved):

davidcassany:
rebased to fix conflicts with 8d69603

ipetrov117 (approved):

atanasdinov (approved):

PR #537, opened 2026-07-21
Allow external API VIP providers

ravan:
Description:

## What this PR does

Adds an optional `apiVIPMode` setting to the Kubernetes network configuration.

- `managed` or omitted preserves the existing behavior, adding MetalLB and endpoint-
  copier-operator.
- `external` keeps the configured API VIP but skips the built-in load-balancer Helm
  charts.
- Validates supported values and documents the new setting.
- Adds tests for managed, external, and invalid modes.

## Why

This allows clusters to use an externally managed load balancer without including the
built-in load-balancing components.

## Testing

- Added configuration tests covering all API VIP modes.
- Focused Go tests passed.
- Both `elemental3` and `elemental3ctl` built successfully in an AMD64 Go 1.26
  container.

### DIFF
--- a/docs/configuration-directory.md
+++ b/docs/configuration-directory.md
@@ -147,6 +147,7 @@ network:
     apiVIP: 192.168.122.100
     apiHost: api.cluster01.example.com
     apiVIP6: fd12:3456:789a::21
+    apiVIPMode: managed
 ```
 
 * `manifests` - Optional; Defines remote Kubernetes manifests to be deployed on the cluster.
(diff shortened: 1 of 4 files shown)


e-minguez:
I'd say it would be good perhaps to also include a "none" mode that doesn't even configure the IP for the apiVIP on the RKE2/k3s config files.

docs/configuration-directory.md
```diff
@@ -170,8 +171,9 @@ network:
...
 * `network`:
-  * `apiVIP` - Required for multi-node clusters if not using `apiVIP6`; Specifies the IPv4 address which will serve as the cluster LoadBalancer, backed by MetalLB.
-  * `apiVIP6` -  Required for multi-node clusters if not using `apiVIP`; Specifies the IPv6 address which will serve as the cluster LoadBalancer, backed by MetalLB.
+  * `apiVIP` - Required for multi-node clusters if not using `apiVIP6`; Specifies the IPv4 address which will serve as the cluster LoadBalancer. By default, setting this field adds MetalLB and endpoint-copier-operator to the image build.
+  * `apiVIP6` -  Required for multi-node clusters if not using `apiVIP`; Specifies the IPv6 address which will serve as the cluster LoadBalancer. By default, setting this field adds MetalLB and endpoint-copier-operator to the image build.
+  * `apiVIPMode` - Optional; Selects how the API VIP is provided. Omitted or `managed` keeps the default built-in MetalLB and endpoint-copier-operator install. `external` keeps `apiVIP` or `apiVIP6` configured but disables adding those built-in Helm charts at image build time.
```
davidcassany, line 176: I am wondering if instead of `managed` shouldn't we name it `metallb`. If in the future we add support for additional or different load balancers probably it will make more sense to name the stack.

internal/config/v0/config.go
```diff
@@ -229,7 +229,7 @@ func parseKubernetes(f vfs.FS, configDir Dir, k *kubernetes.Kubernetes, r *relea
 		return fmt.Errorf("reading config file: %w", err)
 	}
 
-	if k.Network.APIVIP4 != "" || k.Network.APIVIP6 != "" {
+	if k.Network.IsHA() && k.Network.APIVIPMode != "external" {
```
davidcassany, line 232: On the same idea I'd probably make an explicit match against the network mode rather than filter out the `external`.

davidcassany (commented):
Looks good to me, just a couple of nit comments regarding the apiVIPMode values.

docs/configuration-directory.md
atanasdinov: The specific value would read better if we had:
- `loadBalancer`: `metallb | external`

I guess both the above and `apiVIPMode`: `managed | external` are fine.

atanasdinov:
The functionality has been merged as part of https://github.com/SUSE/elemental/pull/578.

PR #539, opened 2026-07-23
Install LCM by default during customize

dharmit:
- Adds LCM during `elemental3 init`
- Mandates LCM for `elemental3 customize`

### DIFF
--- a/examples/elemental/customize/multi-node/release.yaml
+++ b/examples/elemental/customize/multi-node/release.yaml
@@ -7,3 +7,4 @@ components:
     - chart: metallb # from core release
     - chart: rancher # from solution release
       valuesFile: rancher.yaml
+    - chart: elemental-lifecycle-manager
--- a/internal/cli/action/customize.go
+++ b/internal/cli/action/customize.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os/signal"
 	"path/filepath"
+	"slices"
 	"strings"
 	"syscall"
 	"time"
@@ -33,6 +34,7 @@ import (
 	v0 "github.com/suse/elemental/v3/internal/config/v0"
 	"github.com/suse/elemental/v3/internal/customize"
 	"github.com/suse/elemental/v3/internal/image"
+	"github.com/suse/elemental/v3/internal/image/release"
 	"github.com/suse/elemental/v3/pkg/extractor"
 	"github.com/suse/elemental/v3/pkg/helm"
 	"github.com/suse/elemental/v3/pkg/http"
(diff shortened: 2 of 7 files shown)


internal/cli/action/customize.go
```diff
@@ -182,3 +188,19 @@ func digestCustomizeDefinition(f vfs.FS, args *cmdpkg.CustomizeFlags, imagePath
 		Configuration: conf,
 	}, nil
 }
+
+func validateLifecycleManager(conf *image.Configuration) error {
+	const ElementalLifecycleManager = "elemental-lifecycle-manager"
```
ipetrov117:
```suggestion
	const elementalLifecycleManager = "elemental-lifecycle-manager"
```

examples/elemental/customize/multi-node/release.yaml
```diff
@@ -7,3 +7,4 @@ components:
     - chart: metallb # from core release
     - chart: rancher # from solution release
       valuesFile: rancher.yaml
+    - chart: elemental-lifecycle-manager
```
ipetrov117: Can we bring back the removed new line here. It is a good practice for files to terminate with a new line for compatibility with different program processes.

internal/cli/action/customize_test.go
```diff
@@ -0,0 +1,67 @@
...
+	It("fails when Kubernetes is enabled but LCM is missing", func() {
+		conf := image.Configuration{
+			Release: release.Release{Components: release.Components{Kubernetes: &struct{}{}}},
+		}
+		err := validateLifecycleManager(&conf)
+		Expect(err).ToNot(BeNil())
+		Expect(err.Error()).To(Equal(errorMsg))
+	})
+
+	It("fails when not listed under Helm charts", func() {
```
ipetrov117, line 44: How does this differ from the "fails when Kubernetes is enabled but LCM is missing" case?

ipetrov117 (commented):
Can you refresh my memory on what we agreed on for Rancher handling? Did we not want to have a "basic Rancher" check in code so that we do not deploy the core manifest SUC, but rely on Rancher's one?

dharmit:
> Did we not want to have a "basic Rancher" check in code so that we do not deploy the core manifest SUC, but rely on Rancher's one?

We do. But I don't want to open one big PR for every change required on elemental side. What you are referring to is the "soft dependency" aspect we discussed during the calls. It covers SUC as well as cert-manager.

internal/cli/action/customize_test.go
dharmit: In that case `kubernetes` is explicitly enabled by setting `kubernetes: {}`. In this case it isn't.

Reason for having both checks is that, in my observation, `elemental3 customize` requires setting only one of them to enable Kubernetes on the image that gets generated.

davidcassany (approved):

ipetrov117 (approved):

PR #543, opened 2026-07-27
Export core manifest metadata

davidcassany:
This commit exports the core's Metadata objects so they can be declared form the public elemental API.

### DIFF
--- a/pkg/manifest/api/core/manifest.go
+++ b/pkg/manifest/api/core/manifest.go
@@ -31,6 +31,8 @@ type Components = corev1.Components
 type OperatingSystem = corev1.OperatingSystem
 type Kubernetes = corev1.Kubernetes
 type Image = corev1.Image
+type Metadata = corev1.Metadata
+type Elemental = corev1.Elemental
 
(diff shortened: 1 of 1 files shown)


davidcassany:
This will facilitate code such as this [line in LCM](https://github.com/SUSE/elemental-lifecycle-manager/blob/c9783050d9c907671872d5d7fcd7e0eead5fe19b/internal/controller/release_controller_test.go#L78)

pkg/manifest/api/core/manifest.go
```diff
@@ -31,6 +31,8 @@ type Components = corev1.Components
 type OperatingSystem = corev1.OperatingSystem
 type Kubernetes = corev1.Kubernetes
 type Image = corev1.Image
+type Metadata = corev1.Metadata
```
davidcassany, line 34: The Metadata in Solution and Core manifest is no longer the same, so core's Metadata needs to be publicly exported to be declared and initiated in other projects.

ipetrov117 (approved):

atanasdinov (approved):

PR #548, opened 2026-07-30
Do not overlay kubernetes files

davidcassany:
This commit changes the customize behavior to embed generated and copied over files to ignition configuration instead of using the overlay directory tree during the installation.

Before some files were provisioned via Ignition and other via installation overlays. This commit sets a consistent behavior so that all configuration files in /var/lib/elemental are essentially provisioned at the same time via Ignition.

All unpacked k8s files (tarball, install script) are left as overlay as they were.

### DIFF
--- a/internal/butane/config.go
+++ b/internal/butane/config.go
@@ -18,7 +18,10 @@ limitations under the License.
 package butane
 
 import (
+	"errors"
 	"fmt"
+	"io"
+	"io/fs"
 	"path/filepath"
 
 	base "github.com/coreos/butane/base/v0_6"
@@ -29,6 +32,8 @@ import (
 	"github.com/suse/elemental/v3/pkg/sys"
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
 )
+
+const maxResourceSize int64 = 16 * 1024 * 1024
 
 // Config represents a basic butane configuration
 type Config struct {
@@ -47,6 +52,35 @@ func (c *Config) MergeInlineIgnition(ignitionConf string) {
 	merge.Inline = &ignitionConf
 
 	c.Ignition.Config.Merge = append(c.Ignition.Config.Merge, merge)
+}
+
+// AddFileInline adds the given filepath and inline content to the butane config
+func (c *Config) AddFileInline(filename string, data *string, mode fs.FileMode) {
+	c.Storage.Files = append(c.Storage.Files, base.File{
+		Path: filename,
+		Contents: base.Resource{
+			Inline: data,
+		},
+		Mode: toButaneMode(mode),
+	})
+}
+
+// AddFileInlineFromReader adds the given filepath and inline content from the given reader, closes the reader.
+// It does not support data payloads bigger than 16MiB as this is meant for text files
+func (c *Config) AddFileInlineFromReader(filename string, rc io.ReadCloser, mode fs.FileMode) error {
+
+	resource, err := resourceFromText(rc, maxResourceSize)
+	if err != nil {
+		return nil
+	}
+
+	c.Storage.Files = append(c.Storage.Files, base.File{
+		Path:     filename,
+		Contents: resource,
+		Mode:     toButaneMode(mode),
+	})
+
+	return nil
 }
 
 // AddSystemdUnit adds an inline unit object in butane configuration
@@ -98,4 +132,48 @@ func TranslateBytes(s *sys.System, butane any) ([]byte, error) {
 	}
 	s.Logger().Debug("Butane configuration translated:\n--- Generated Ignition Config ---\n%s", string(ignitionBytes))
 	return ignitionBytes, nil
+}
+
+// toButaneMode converts fs.FileMode into *int
+func toButaneMode(fm fs.FileMode) *int {
+	// Start with the standard 9 permissions
+	mode := uint32(fm.Perm())
+
+	// Add special bits back if they exist in the fs.FileMode
+	if fm&fs.ModeSetuid != 0 {
+		mode |= 04000
+	}
+	if fm&fs.ModeSetgid != 0 {
+		mode |= 02000
+	}
+	if fm&fs.ModeSticky != 0 {
+		mode |= 01000
+	}
+
+	modeInt := int(mode)
+	return &modeInt
+}
+
+// resourceFromText is meant for plain text files, closes the reader
+func resourceFromText(rc io.ReadCloser, maxBytes int64) (_ base.Resource, err error) {
+	defer func() {
+		err = errors.Join(err, rc.Close())
+	}()
+
+	// Limit to maxBytes + 1 so we can detect if it overflows
+	lr := io.LimitReader(rc, maxBytes+1)
+
+	data, err := io.ReadAll(lr)
+	if err != nil {
+		return base.Resource{}, err
+	}
+
+	// If we read more than maxBytes, the source was too large
+	if int64(len(data)) > maxBytes {
+		return base.Resource{}, fmt.Errorf("input exceeds maximum allowed size of %d bytes", maxBytes)
+	}
+
+	return base.Resource{
+		Inline: new(string(data)),
+	}, nil
 }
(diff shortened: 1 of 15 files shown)


internal/butane/config.go
```diff
@@ -30,6 +33,8 @@ import (
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
 )
 
+const maxResourceSize int64 = 16 * 1024 * 1024
```
davidcassany, line 36: This is the maximum size permitted for the file contents added to ignition configuration when they are user provided. As the file needs to be entirely loaded in memory and there are no sanity checks on the content I opted to add a limitation here. This is mostly for remote files and manifests in config, if someone adds a link to huge file or a big binary in config it could easily cause a run out of memory. This is to prevent it.

Since this is only used for inline text yaml files I think 16MiB is huge enough.

internal/config/helm.go
```diff
@@ -95,11 +91,7 @@ func (h *Helm) Configure(conf *image.Configuration, rm *resolver.ResolvedManifes
 	return chartFiles, helmSecrets, nil
```
ipetrov117: Now that we have included the butane config to the `helm.Configure` function, I wonder whether it would make more sense to handle the `helmSecrets` map directly as part of the `helm.Configure` function as opposed to handling them in the m.setupManifests function as [additionalManifests](https://github.com/davidcassany/elemental/blob/update_kubernetes_config_management/internal/config/kubernetes.go#L94). 

Perhaps having a `writeHelmSecrets` function that does what `h.createHelmSecretFileMap` does, but adds to ignition as well. That way we would handle all helm-relevant resource configurations as part of the `helm.Configure` function and not have to rely on outside functions to setup the secrets themselves. Furthermore, by doing this we could effectively drop the [additionalManifests](https://github.com/davidcassany/elemental/blob/update_kubernetes_config_management/internal/config/kubernetes.go#L156-L159) logic in the `setupManifests` function, as it was handling only the helm secrets and currently we do not offer any other additional resources for it. Let me know what you think.

internal/config/manager.go
```diff
@@ -144,7 +139,14 @@ func (m *Manager) ConfigureComponents(ctx context.Context, conf *image.Configura
 		}
 	}
 
-	if err = m.configureIgnition(conf, output, k8sScript, k8sConfScript, extensions); err != nil {
+	if isKubernetesEnabled(conf) {
+		err = m.unpackKubernetesArtifacts(ctx, rm, output)
```
ipetrov117: nit: I wonder whether it would make sense to move this directly before the [m.configureKubernetes](https://github.com/davidcassany/elemental/blob/update_kubernetes_config_management/internal/config/ignition.go#L88) call under the `configureSystem` function. That way we would not have to double the `isKubernetesEnabled` check.

pkg/http/download.go
```diff
@@ -28,29 +28,30 @@ import (
 	"github.com/suse/elemental/v3/pkg/sys/vfs"
 )
 
-func DownloadFile(ctx context.Context, fs vfs.FS, url, path string) error {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return fmt.Errorf("creating request: %w", err)
-	}
+type Downloader interface {
```
ipetrov117: To keep the consistency of defining interfaces closer to where they are actually referenced, would it make sense to move this interface under the manager itself? Then here we would have a variable that points to an empty `downloader` struct which is referred in the manager. WDYT?

ipetrov117 (commented):
Overall it looks good to me. Left a couple of small comments, will be happy to approve once we discuss them. Thanks!

internal/config/manager.go
davidcassany: We can consider that, in fact, before `configureSystem` was named `configureIgnition` (IMHO this was already a poor name as it was not configuring Ignition itself, it was creating Ignition configuration) and it simply felt so wrong unpacking k8s inside this method. Later on I renamed the method as I didn't like it anyways and did not look back to re-arrange k8s unpack. I am fine moving it back inside the configureSystem.

internal/config/kubernetes.go
```diff
@@ -215,50 +246,153 @@ func writeK8sConfigDeployScript(fs vfs.FS, output Output, k kubernetes.Kubernete
...
+		InitHostname         string
+	}{
+		KubernetesDir:        filepath.Dir(deployScript),
+		ManifestDeployScript: deployScript,
+		InitHostname:         initHostname,
 	}
 
-	fullPath := filepath.Join(destDir, k8sConfDeployScriptName)
-	relativePath := filepath.Join(relativeK8sPath, k8sConfDeployScriptName)
+	data, err := template.Parse(k8sResourcesUnitName, k8sResourceUnitTpl, &values)
```
dharmit, line 271: Like we have moved some constants from `ignition.go`, would it make sense to move `k8sResourceUnitTpl` variable too as it's being called in this file?

pkg/http/download.go
davidcassany: We have both approaches in the code currently. I guess this is more a matter of taste, each approach has its pros and both could be applicable at the same time.

The rationale with the current code is that the `downloader` can't be manually initiated outside it's own package as it is not exported. An `http.NewDownloader()` is required which returns the public interface. This pattern ensures methods from `http` package are always used through an interface and never called form a specific instance type.

IMO the pattern of moving the interface definition inside the manager provides a couple of benefits:
* filterout unneeded methods (e.g. io.ReadCloser vs io.Reader)
* entirely decouple manager from the interface implementation, only if the interface in not initiated within the manager itself (no `http` import required in manager).

Anyway I am fine changing it.

ipetrov117: What I meant, was that currently all other interfaces that the manager refers to are defined within the `config` package, so I wondered whether keeping that consistency made sense. That said, I recognise that this falls much to preference, so I am fine keeping the changes as they are. (Should have probably marked this comment as a "nit").

internal/config/manager.go
ipetrov117: Yes, the naming was unfortunate, I too think that `configureSystem` better matches the function's objectives. I suggested the move as it would bring the K8s artefact preparation step closer the the configuration itself and will avoid a double K8s validation, but I also see the use-case where we can overload the `configureSystem` with non-configuration code as well. Anyway, I am fine with both approaches for the moment.

internal/config/kubernetes.go
```diff
@@ -215,50 +246,153 @@ func writeK8sConfigDeployScript(fs vfs.FS, output Output, k kubernetes.Kubernete
...
+	if err != nil {
+		return "", fmt.Errorf("parsing config script template: %w", err)
```
dharmit: nit: perhaps a different message as `parsing config script template` is used both here and in `generaeteK8sConfigUnit` function as well. I could be wrong and it might be perfect to have same messages at both places, so correct me if that's the case. :)

dharmit (approved):
Two small nits. LGTM otherwise.

internal/config/kubernetes.go
```diff
@@ -23,127 +23,148 @@ import (
...
-		return "", "", fmt.Errorf("writing kubernetes config deployment script: %w", err)
+		return fmt.Errorf("generating RKE2 configuration: %w", err)
 	}
 
-	return k8sResourceScript, k8sConfScript, nil
+	return nil
 }
 
-func (m *Manager) setupManifests(ctx context.Context, k *kubernetes.Kubernetes, additionalManifests map[string][]byte, output Output) (string, error) {
+func (m *Manager) setupManifests(ctx context.Context, k *kubernetes.Kubernetes /*additionalManifests map[string][]byte,*/, butaneCfg *butane.Config) (string, error) {
```
ipetrov117: I am guessing we are leaving the additionalManfiests parameter commented to reflect the possible future functionality enhancement?

ipetrov117 (approved):
Looks good to me!

internal/config/kubernetes.go
davidcassany: uh... this is a leftover of my trials when fixing tests 😞

davidcassany: Gonna remove this comment and rebase and squash everything before merging

ipetrov117 (approved):

PR #551, opened 2026-08-05
Error on unknown hostname

frelon:
During kubernetes provisioning the current behavior is to assume a node
is a 'server' and if found in the list of nodes we set it to the found
node-type.

In this commit we abort with an error if the node is not found for
multi-node clusters.

### DIFF
--- a/internal/config/kubernetes_test.go
+++ b/internal/config/kubernetes_test.go
@@ -345,6 +345,26 @@ var _ = Describe("Kubernetes", func() {
 			Expect(confScript).ToNot(BeEmpty())
 		})
 
+		It("Defaults to a server node for a single-node cluster with no declared nodes", func() {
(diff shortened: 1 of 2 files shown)


davidcassany (approved):

dharmit:
EDIT: This is probably a me problem. I'm facing this with main branch as well. Running `openssl passwd -6` as mentioned in the code comment in butane.yaml and using the hash it provided also doesn't let me login to the VM. Weird. 

I don't know if it's just me, but a multi-node or single-node image built using this PR is not letting me login with root-linux credentials that we have configured in butane.yaml for these examples. 

I do make a change before building the image for both example types which is rebase against main to have changes from #549 pulled in so that customize doesn't fail with LCM requirement. But I don't suppose that should lead to such a problem.

PR #553, opened 2026-08-06
Adds Installer information to release-info output

dharmit:
This helps people use `release-info` output to download the required
images before running `customize` with `--local` flag as the image
referred to by Installer is used when building a VM using `customize`.

Example output:
```sh
# core manifest
$ cat /tmp/core/release_manifest.yaml | yq '.components.operatingSystem.image'
base: registry.suse.com/elemental/base-os-kernel-default:16.0-3.10
iso: registry.suse.com/elemental/base-os-kernel-default-iso:16.0-3.11

# before
$ sudo ./build/elemental3 release-info ./examples/elemental/customize/multi-node/suse-solution-manifest.yaml | grep -A2 "Operating System"
│ Operating System          │ SLES 16.0      │ registry.suse.com/elemental/base-os-kernel-default:16.0-3.16 │
│ Kubernetes                │ v1.35.6+rke2r1 │ registry.suse.com/elemental/rke2/rke2-tar:1.35.6_rke2r1-9.1  │

# after
$ sudo ./build/elemental3 release-info ./examples/elemental/customize/multi-node/suse-solution-manifest.yaml | grep -A2 "Operating System"
│ Operating System          │ SLES 16.0      │ registry.suse.com/elemental/base-os-kernel-default:16.0-3.16     │
│ Installer                 │ SLES 16.0      │ registry.suse.com/elemental/base-os-kernel-default-iso:16.0-3.15 │
│ Kubernetes                │ v1.35.6+rke2r1 │ registry.suse.com/elemental/rke2/rke2-tar:1.35.6_rke2r1-9.1      │
```

### DIFF
--- a/internal/cli/action/release_info.go
+++ b/internal/cli/action/release_info.go
@@ -271,6 +271,15 @@ func printInfraData(cm *core.ReleaseManifest, out io.Writer) error {
 		data = append(data, []string{"Operating System", osVersion, cm.Components.OperatingSystem.Image.Base})
 	}
 
+	if cm.Components.OperatingSystem.Image.ISO != "" {
+		installerVersion := "Unknown"
+		parts := strings.Split(cm.Components.OperatingSystem.Image.ISO, ":")
+		if len(parts) > 1 {
+			installerVersion = "SLES" + " " + strings.Split(parts[1], "-")[0]
+		}
+		data = append(data, []string{"Installer", installerVersion, cm.Components.OperatingSystem.Image.ISO})
+	}
+
 	if cm.Components.Kubernetes != nil {
 		data = append(data, []string{"Kubernetes", cm.Components.Kubernetes.Version, cm.Components.Kubernetes.Image})
 	}
--- a/internal/cli/action/release_info.go
+++ b/internal/cli/action/release_info.go
@@ -268,8 +268,17 @@ func printInfraData(cm *core.ReleaseManifest, out io.Writer) error {
 		if len(parts) > 1 {
 			osVersion = "SLES " + strings.Split(parts[1], "-")[0]
 		}
-		data = append(data, []string{"Operating System", osVersion, cm.Components.OperatingSystem.Image.Base})
 	}
+	data = append(data, []string{"Operating System", osVersion, cm.Components.OperatingSystem.Image.Base})
+
+	installerVersion := "Unknown"
+	if cm.Components.OperatingSystem.Image.ISO != "" {
+		parts := strings.Split(cm.Components.OperatingSystem.Image.ISO, ":")
+		if len(parts) > 1 {
+			installerVersion = "SLES" + " " + strings.Split(parts[1], "-")[0]
+		}
+	}
+	data = append(data, []string{"Installer", installerVersion, cm.Components.OperatingSystem.Image.ISO})
 
 	if cm.Components.Kubernetes != nil {
 		data = append(data, []string{"Kubernetes", cm.Components.Kubernetes.Version, cm.Components.Kubernetes.Image})
--- a/internal/cli/action/release_info_test.go
+++ b/internal/cli/action/release_info_test.go
@@ -102,6 +102,7 @@ components:
 		Expect(buffer).To(ContainSubstring("CORE PLATFORM"))
 		Expect(buffer).To(ContainSubstring("suse-core-test"))
 		Expect(buffer).To(ContainSubstring("registry.suse.com/elemental/base-os-kernel-default:16.0-2.4"))
+		Expect(buffer).To(ContainSubstring("registry.suse.com/elemental/base-os-kernel-default-iso:16.0-2.7"))
 
 		Expect(buffer).To(ContainSubstring("INFRASTRUCTURE COMPONENTS"))
 		Expect(buffer).To(ContainSubstring("SLES 16.0"))


ipetrov117:
While this is true, when running `cusomitze`, the opposite is true, when doing an upgrade. The `base-os-kernel-default-iso` image is only used during initial image creation, after that we always refer and use the `base-os-kernel-default`. Furthermore, the `base-os-kernel-default` is the actual image that represents the base OS state, `base-os-kernel-default-iso` is just a wrapper that ships an installer ontop. 

Since the field is labeled `Operating System`, I'd suggest to keep it pointing to the `base-os-kernel-default`, as that is the actual image that is used for the base OS. Both during `customize` (through the `base-os-kernel-default-iso`) and upgrade. If we put `base-os-kernel-default-iso` under `Operating System`, we would be inviting questions such as - "Why is it called *-iso, what do I need to use for raw?" etc..

Alternatively, we could provide the `base-os-kernel-default-iso` reference in a different field (e.g. `Installer`), but I am not sure we would get much out of it. I'll let others chime in as well, but IMO the `release-info` command should provide easy to follow and understand information about the general system that this manifest represents - what OS version it ships (e.g. `base-os-kernel-default` build), what RKE2 version it ships, which charts it ships. Going further into detail about other images that we use to achieve this basic state, could just cause confusion.

dharmit:
@ipetrov117 thank you for the explanation. 👍🏽 

> Alternatively, we could provide the `base-os-kernel-default-iso` reference in a different field (e.g. `Installer`), but I am not sure we would get much out of it.

I'd prefer this. Personally, my workflow is:
- Do `elemental3 release-info`
- Store container images in its output into a shell variable
- Download the images
- Use `--local` flag when running `customize`

I don't see a reason right now, but at some point I feel it'll be valuable to add a flag to `release-info` command that prints only the container images. These output can be used to download all required images before running `customize`. Would be very useful for folks relying on `--local`.

davidcassany (commented):
I am also unsure about this change, I think the `Operating System` should point to the OS image. However I see the benefits of having the ISO listed too, so I think the ISO deserves another field or making the `Operator System` a multi value stanza. I am not sure what would be the best approach. IMHO release info should give the references of both, the installer image and the OS image.

atanasdinov:
Agree with both Ivo & David here. Adding `Installer` field is fine, but keep in mind the behaviour for the `--local` flag has been already on the radar to adjust with the changes around airgap functionality that @dbw7 is looking forward to introduce.

dharmit:
Updated the PR such that `release-info` output now incldues "Installer" field and relevant value. Also the PR description is updated with example output. PTAL again folks. :)

internal/cli/action/release_info.go
```diff
@@ -271,6 +271,15 @@ func printInfraData(cm *core.ReleaseManifest, out io.Writer) error {
...
 	}
 
+	if cm.Components.OperatingSystem.Image.ISO != "" {
+		installerVersion := "Unknown"
+		parts := strings.Split(cm.Components.OperatingSystem.Image.ISO, ":")
+		if len(parts) > 1 {
+			installerVersion = "SLES" + " " + strings.Split(parts[1], "-")[0]
+		}
+		data = append(data, []string{"Installer", installerVersion, cm.Components.OperatingSystem.Image.ISO})
+	}
```
atanasdinov, line 280: Feels misconstructed to me. Initialising the value and setting it should happen outside of the conditional.
```suggestion
	installerVersion := "Unknown"
	if cm.Components.OperatingSystem.Image.ISO != "" {
		parts := strings.Split(cm.Components.OperatingSystem.Image.ISO, ":")
		if len(parts) > 1 {
			installerVersion = "SLES" + " " + strings.Split(parts[1], "-")[0]
		}
	}
	data = append(data, []string{"Installer", installerVersion, cm.Components.OperatingSystem.Image.ISO})
```

internal/cli/action/release_info.go
```diff
@@ -271,6 +271,15 @@ func printInfraData(cm *core.ReleaseManifest, out io.Writer) error {
 		data = append(data, []string{"Operating System", osVersion, cm.Components.OperatingSystem.Image.Base})
 	}
 
+	if cm.Components.OperatingSystem.Image.ISO != "" {
+		installerVersion := "Unknown"
+		parts := strings.Split(cm.Components.OperatingSystem.Image.ISO, ":")
+		if len(parts) > 1 {
+			installerVersion = "SLES" + " " + strings.Split(parts[1], "-")[0]
```
atanasdinov, line 278: Not sure how to approach this version. This is the installer which is based on SLES 16, not SLES 16 itself. Any thoughts?

dharmit: I think we should leave it blank using a hyphen in that case.

```
│ Installer                 │ -      │ registry.suse.com/elemental/base-os-kernel-default-iso:16.0-3.15 │
```
Or hardcode it to N/A.
```
│ Installer                 │ N/A      │ registry.suse.com/elemental/base-os-kernel-default-iso:16.0-3.15 │
```

dharmit: @atanasdinov wdyt here?

internal/cli/action/release_info.go
dharmit: Resolved. In the block above this one as well.

internal/cli/action/release_info.go
dharmit: As discussed, letting this be SLES 16.0 for the time being as keeping it empty would not help either.

atanasdinov (approved):

ipetrov117 (approved):

PR #554, opened 2026-08-06
WIP: Three way merge

davidcassany:
This is just a draft PR to debug why the former PR #508 is not passing integration tests. Before squashing and rebasing tests passed on former PR, wondering if  a new clean project build in OBS will help.

### DIFF
--- a/pkg/merge/merge.go
+++ b/pkg/merge/merge.go
@@ -1,0 +1,191 @@
+/*
+Copyright © 2025-2026 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


ipetrov117:
@davidcassany, I think #508's failure seems like it is something sporadic related to the PR's testing environment. I tried to reproduce this locally with mixed success. IMO if the tests pass on the clean OBS project, we are safe to merge this.

davidcassany:
> @davidcassany, I think #508's failure seems like it is something sporadic related to the PR's testing environment. I tried to reproduce this locally with mixed success. IMO if the tests pass on the clean OBS project, we are safe to merge this.

Yes, I think we can close this PR and merge #508, this was been around quite a while and the OBS project might have some leftover I can't easily see.

PR #556, opened 2026-08-10
Fix systemd-repart discard hanging on large disks

### LINKED ISSUE #555: Elemental Failing on Large Disks
The current usage of systemd-repart causes certain disks to wait for a full zero-write of the entire disk before proceeding with the installation. On a large disk, this can result in the installation hanging for a while and eventually timing out.

dbw7:
Fixes #555 by doing an explicit discard of the entire disk all at once. This discard is a lazy discard that allows the installation to proceed and then lets the disk's native TRIM engine trim the disk on its own time. Installation proceeds if a disk doesn't support `blkdiscard` but installation is aborted if any other error comes up like "disk in use"

### DIFF
--- a/pkg/repart/disk_repart.go
+++ b/pkg/repart/disk_repart.go
@@ -20,9 +20,11 @@ package repart
 import (
 	_ "embed"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


atanasdinov (approved):
Let's get this merged, but we'd ideally need to consult @davidcassany once he's available.

ipetrov117 (approved):

davidcassany (approved):
Looks good to me

PR #559, opened 2026-08-18
Ignore soft dependencies of LCM and chart prioritization

dharmit:
Ignore soft dependencies, i.e., SUC and cert-manager, of LCM during `elemental3 customize`.

Also, prioritze charts in core manifest over the one in solution manifest.


internal/config/lcm.go
```diff
@@ -0,0 +1,127 @@
...
+
+const (
+	elementalLifecycleManager = "elemental-lifecycle-manager"
+	rancher                   = "rancher"
+	systemUpgradeController   = "system-upgrade-controller"
+	certManager               = "cert-manager"
+)
+
+// lcmWebhookValues is used to marshal data from values file for LCM chart
+type lcmWebhookValues struct {
```
ipetrov117: I do not foresee this type being every used outside of `evaluateLCMDeps`, as such would it make sense to include it as an anonymous struct inside the func?

Also we could maybe simplify the struct itself to something like:

```golang
var lcmWebhookValues struct {
		Webhook struct {
			Cert *struct {
				CreateDefault bool `yaml:"createDefault"`
			} `yaml:"cert"`
		} `yaml:"webhook"`
	}
```

And then check against  the Cert struct pointer and the CreateDefault field themselves. That would be enough, as if the user has continuously provided the "cert:" field with a non-true `createDefault`, then it is up to them to define the other fields, in our eyes this should be enough for the trigger to disable the cert-manager dependency.

internal/config/lcm.go
```diff
@@ -0,0 +1,127 @@
...
+			ExistingSecret string `yaml:"existingSecret"`
+			CABundle       string `yaml:"caBundle"`
+		} `yaml:"cert"`
+	} `yaml:"webhook"`
+}
+
+// evaluateLCMDeps removes dependencies of LCM if they are satisfied separately, i.e.,
+// - if Rancher chart is enabled in release.yaml, it removes dependency on system-upgrade-controller
+// - if the values files contains custom certificate configuration, it removes dependency on cert-manager
+func evaluateLCMDeps(enabled []release.HelmChart, corePlatform *core.ReleaseManifest, valueFiles map[string]string, valuesResolver helmValuesResolver) error {
```
ipetrov117: A couple of suggestions here:
1. I am not sure that the `valueFiles map[string]string` parameter is needed here. The values file location can be taken from the LCM entry in the `enabled []release.HelmChart` slice.
2. This function seems to be doing too much, could we perhaps split out some of its logic in separate functions. For example:
   - We could extract the logic to find a chart from the core manifest in a separate generic function that could be reused for other use-cases in the future.
   - We could extract the logic to resolve a specific charts values (or a portion of it)  in a separate generic function that could then be reused for other use-cases in the future.

internal/config/lcm.go
```diff
@@ -0,0 +1,127 @@
...
+	var lcmChart *api.HelmChart
+
+	var (
+		lcmEnabled     = false
+		rancherEnabled = false
+	)
+	for _, chart := range enabled {
+		if chart.Name == rancher {
+			rancherEnabled = true
```
ipetrov117: This may be a personal preference, but instead of populating boolean values, would it make sense to convert the enabled slice to a map and check against the map keys. That way we could do the dependency removal in the same block that we check whether the chart is enabled.

internal/config/lcm.go
```diff
@@ -0,0 +1,127 @@
...
+			continue
+		}
+		if chart.Name == elementalLifecycleManager {
+			lcmEnabled = true
+		}
+	}
+	if !lcmEnabled {
+		// nothing to do!
+		return nil
```
ipetrov117: Shouldn't we error out here? IIRC, we agreed that LCM is a required chart that should be specified by the user under `release.yaml`. By returning nil here, we do not have any gate that ensures that LCM will be on the cluster. Probably missing something.

internal/config/lcm.go
```diff
@@ -0,0 +1,127 @@
...
+		if chart.GetName() == elementalLifecycleManager {
+			lcmChart = chart
+			break
+		}
+	}
+
+	if lcmChart == nil {
+		// this could be the case if using core manifest that doesn't contain LCM charts which is the case currently
+		// TODO (dharmit): remove this check once the core manifest includes LCM chart by default
+		return nil // nothing to do!
```
ipetrov117: Shouldn't we also error out here as well? My main thinking is that LCM will always be a part of the core manifest, so it not being present in the parsed `corePlatform` struct should be flagged. 

Or did you want to make the `elemental3` binary backwards compatible? If that is the case, I guess we need to decide how we want to handle this. IMO if LCM is going to be an expected component in our setup we need to be explicit here and guard against a missing LCM in `release.yaml` as well as in the core manifest. Let me know what you think.

internal/config/lcm.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if rancherEnabled {
+		// remove system-upgrade-controller from list of dependencies
+		for i, dep := range lcmChart.DependsOn {
+			if dep.Name == systemUpgradeController {
+				lcmChart.DependsOn = slices.Delete(lcmChart.DependsOn, i, i+1)
```
ipetrov117: This may be a personal preference, so feel free to ignore, but would it make sense to use the `slices.DeleteFunc` here and instead of this whole block, have something like:

```golang
lcmChart.DependsOn = slices.DeleteFunc(lcmChart.DependsOn,
			func(d api.HelmChartDependency) bool { return d.Name == systemUpgradeController })
```

internal/config/lcm.go
ipetrov117, line 1: Since this file consists of a single function and is doing essentially a helm chart configuration, would it make sense to merge its logic and its `_test.go` with the `helm.go` file? That way we would have all the chart handling logic at a single place and will keep the number of files for this package to a minimum.

ipetrov117 (commented):
Good job on this. I left a couple of suggestions/comments, please let me know what you think.

internal/config/lcm.go
dharmit: No, backwards compatibility isn't on my mind here. This and the https://github.com/SUSE/elemental/pull/559#discussion_r3861806431 check are solely for the reason mentioned in the code comment here.

The current latest core manifest doesn't have LCM and its CRD charts in it. Once that is addressed, I'll remove these checks. Also, we can un-revert #549 once this is addressed.

ipetrov117: Sure, but what I am asking is "isn't it enough to have these two errors here" to do the actual validation and not have to keep the validation logic from #549 at all?

Also, it may be a personal preference, but adding code just so we remove it later seems weird to me. Especially if, IIUC, adding these two errors here would mitigate the need for the validation portion of #549. I may be missing something, though.

dharmit: While there's nothing really wrong, the way it's done in #539 is quite early in the flow before creation of the temporary directory for build artifacts.

Otherwise, I don't disagree with what you are suggesting.

internal/config/lcm.go
dharmit: Makes sense. Done.

internal/config/lcm.go
dharmit: > I am not sure that the `valueFiles map[string]string` parameter is needed here. The values file location can be taken from the LCM entry in the `enabled []release.HelmChart` slice.

Thanks for pointing. Removed. IMO, the evaluateDeps method has now been simplified. PTAL.

> We could extract the logic to find a chart from the core manifest in a separate generic function that could be reused for other use-cases in the future.

Ack. Done.

> We could extract the logic to resolve a specific charts values (or a portion of it)  in a separate generic function that could then be reused for other use-cases in the future.

This one took me a while. But PTAL now and let me know what you think.

internal/config/lcm.go
dharmit: Modified.

internal/config/lcm.go
dharmit: I moved things to `helm.go` and `helm_test.go` because of changing it from function to method.

internal/config/lcm.go
dharmit: Done. Thanks for pointing.

davidcassany (approved):

internal/config/helm.go
```diff
@@ -167,6 +179,104 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
...
+	for _, chart := range enabled {
+		enabledMap[chart.Name] = struct{}{}
+	}
+	if _, ok := enabledMap[elementalLifecycleManager]; !ok {
+		// LCM isn't enabled in release.yaml; nothing to do!
+		// TODO (dharmit): remove this check once the core manifest includes LCM chart by default
+		return nil
+	}
+
+	if _, ok := enabledMap["rancher"]; ok {
```
dbw7: Can we swap "rancher" with the rancher variable at the top?

internal/config/helm.go
```diff
@@ -167,6 +179,104 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
...
+		// remove system-upgrade-controller from list of dependencies
+		for _, dep := range lcmChart.DependsOn {
```
dbw7: I think Ivo's suggestion actually allows you to turn

```
for _, dep := range lcmChart.DependsOn {
	if dep.Name == systemUpgradeController {
		lcmChart.DependsOn = slices.DeleteFunc(lcmChart.DependsOn, func(dependency api.HelmChartDependency) bool {
			return dependency.Name == systemUpgradeController
		})
		break
	}
}
```

Into just
```
lcmChart.DependsOn = slices.DeleteFunc(lcmChart.DependsOn, func(dependency api.HelmChartDependency) bool {
			return dependency.Name == systemUpgradeController
})
```

internal/config/helm.go
```diff
@@ -167,6 +179,104 @@ func (h *Helm) retrieveHelmCharts(rm *resolver.ResolvedManifest, conf *image.Con
...
+			} `yaml:"webhook"`
+		}
+		err = yaml.Unmarshal(values, &lcmValues)
+		if err != nil {
+			return err
+		}
+
+		if lcmValues.Webhook.Cert.CreateDefault != nil && !*lcmValues.Webhook.Cert.CreateDefault {
+			// checking only createDefault and existingSecret because caBundle could be an empty string
+			for _, dep := range lcmChart.DependsOn {
```
dbw7: I believe the for loop + the if can be dropped here as well just like above so it just turns into:
```
if lcmValues.Webhook.Cert.CreateDefault != nil && !*lcmValues.Webhook.Cert.CreateDefault {
    lcmChart.DependsOn = slices.DeleteFunc(lcmChart.DependsOn, func(dependency api.HelmChartDependency) bool {
			return dependency.Name == certManager
	})
}
```

dbw7 (changes requested):
Just a few small changes, otherwise looks good!

internal/config/helm.go
dharmit: Thanks for correcting me here. 👍🏽

dbw7 (approved):
LGTM

dharmit:
Damn, forgot to rebase before merging. 😢

PR #579, opened 2026-09-04
Add examples and bootstrap script for incoming AWS feature

ipetrov117:
> Important: Merge after #578.

This PR adds configuration examples for the AWS feature introduced as part of #578 and is meant to shed some light in how a user can spin up the necessary test infrastructure and use a customized image to deploy either a single or multi node UC cluster on AWS.

The examples are spit in two groups:
1. `examples/elemental/customize/aws` - a configuration directory that can be used as a base to spin up a customized image.
2. `examples/elemental/runtime-configs` - butane configurations that can be converted to ignition configs. Represent the runtime configurations that can be done over an EC2 instance depending on the desired cluster type (e.g. single/multi).

In addition to the examples, a new `uc_aws_bootstrapper.sh` script is introduced - this script works in 3 phases and is capable of setting up all the necessary AWS resources automatically, so that the user never has to go and manually touch AWS, or know AWS semantics. The script is intended mainly for development purposes, so that we can have a unified way of testing the AWS feature. That said, it can also be used to demo the UC on AWS feature, as well. Also it can further be extended to offer a full "blackbox" deployment approach in the future. 

How would a development workflow look like with this approach:
1. Implement changes to elemental.
2. Run `/uc_aws_bootstrapper.sh bootstrap-infra --allow-ssh` - this will spin up:
    - Necessary static IPs for a 3 cp and 1 wk cluster setup (the number of node IPs is configurable).
    - Necessary security groups for NLBs and nodes.
    - Necessary target groups for RKE2 API, Supervisor and ingress HTTP and HTTPS communications
    - The NLBs themselves.
    - At the end the command will give the static IP and host for the NLB for RKE2 and the public IP for the NLB for application communication.
3. Update [apiVIP](https://github.com/ipetrov117/elemental/blob/aws-examples/examples/elemental/customize/aws/kubernetes/cluster.yaml#L2) and [apiHost](https://github.com/ipetrov117/elemental/blob/aws-examples/examples/elemental/customize/aws/kubernetes/cluster.yaml#L3) with the provided from (2) static IP and host for the RKE2 NLB.
4. Update the Rancher example's [host](https://github.com/ipetrov117/elemental/blob/aws-examples/examples/elemental/customize/aws/kubernetes/helm/values/rancher.yaml#L1) to `rancher.<public_NLB_IP>.sslip.io` - using `sslip.io` here mainly so that we do not have to setup a Route53 hosted zone. 
    > Note: If running on a free AWS account, you may also want to reduce the replicas to 1 and disable some features (e.b. `features: "continuous-delivery=false,harvester=false,managed-system-upgrade-controller=false,turtles=false,uiextension=false,ui-sql-cache=false"`) so that Rancher could start. This is needed, as the largest machine you can use on a free AWS account is 2 vCPI and 4 GB RAM which is way below Rancher's minimum requirements, but for demo purposes the above configurations do the trick.
5. Run `./uc_aws_bootstrapper.sh bootstrap-ami --customized-img <customized.raw>` - this will:
    - Create an S3 bucket if one does not already exist.
    - Publish the image to said bucket.
    - Create the necessary IAM roles and policies so that an EBS snapshot can be created from this bucket's contents.
    - Register an AMI that can be used for EC2 instance spinning.
6. Based on what cluster you want either go to `examples/elemental/runtime-configs/single-node` or `examples/elemental/runtime-configs/multi-node` and run:
    ```
    # Single node example, for multi node run the command for each directory
    # and make sure that for each directory you produce a unique file.
    butane --strict --pretty --files-dir . butane.yaml > "config.ign"
    ```
7. Move generated ignition config(s) to a directory - example:
   ```
   .
   ├── config1.ign 
   ├── config2.ign
   ├── config3.ign 
   └── config4.ign 
   ```
8. Run `./uc_aws_bootstrapper.sh bootstrap-cluster --ignition-config-dir <ignition_dir>` - This will:
    - Lauch 4 EC instances
        > Note: the number of instances is configurable, default is 4. Make sure the number of instances is not larger than the number of statically defined IPs.
    - Wait for the API, Supervisor, HTTP and HTTPS target group's targets to become healthy
9. When done testing, run `./uc_aws_bootstrapper.sh bootstrap-cluster teardown` - This will remove all created infrastructure, ensuring minimal cost.

As the setup is complex, and we are not yet aligned, I feel this script will be useful, both to iterate over and use as-is. However, if you disagree, I am happy to not include it upstream as well. 

Also, real documentation, showcasing all of the above examples will come only after we align about the approach.

### DIFF
--- a/examples/elemental/customize/aws/suse-solution-manifest.yaml
+++ b/examples/elemental/customize/aws/suse-solution-manifest.yaml
@@ -1,0 +1,51 @@
+metadata:
+  name: "suse-solution"
+  version: "4.2.0"
+  creationDate: "2025-12-08"
+corePlatform:
+  # Registry path to the release manifest OCI image of the Core Platform that this SUSE Solution extends
+  # Note: AWS feature is not yet available on registry.suse.com, hence the referral to registry.suse.de
+  image: "registry.suse.de/devel/unifiedcore/main/totest/containers/beta/uc/rke2/rke2-manifest:1.35.6"
+components:
+  systemd:
+    extensions:
+      - name: "longhorn"
+        image: "registry.suse.com/elemental/longhorn:4.69-4.3"
+  helm:
+    charts:
+      - chart: "cert-manager"
+        version: "v1.17.2"
+        namespace: "cert-manager"
+        repository: "jetstack"
+        values:
+          crds:
+            enabled: true
+      - chart: "rancher"
+        version: "2.14.0"
+        namespace: "cattle-system"
+        repository: "rancher"
+        values:
+          replicas: 1
+        dependsOn:
+          - name: "cert-manager"
+            type: "helm"
+      - chart: "longhorn-crd"
+        version: "107.0.0+up1.9.1"
+        namespace: "longhorn-system"
+        repository: "rancher-charts"
+      - chart: "longhorn"
+        version: "107.0.0+up1.9.1"
+        namespace: "longhorn-system"
+        repository: "rancher-charts"
+        dependsOn:
+          - name: "longhorn-crd"
+            type: "helm"
+          - name: "longhorn"
+            type: "sysext"
+    repositories:
+      - name: "jetstack"
+        url: "https://charts.jetstack.io"
+      - name: "rancher"
+        url: "https://releases.rancher.com/server-charts/latest"
+      - name: "rancher-charts"
+        url: "https://charts.rancher.io"
--- a/examples/elemental/customize/aws/butane.yaml
+++ b/examples/elemental/customize/aws/butane.yaml
@@ -1,0 +1,7 @@
+version: 1.6.0
+variant: fcos
+passwd:
+  users:
+  - name: root
+    # Hash for 'linux' passwd created with "openssl passwd -6"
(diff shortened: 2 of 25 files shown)


examples/elemental/customize/aws/suse-solution-manifest.yaml
```diff
@@ -0,0 +1,51 @@
+metadata:
+  name: "suse-solution"
+  version: "4.2.0"
+  creationDate: "2025-12-08"
+corePlatform:
+  # Registry path to the release manifest OCI image of the Core Platform that this SUSE Solution extends
+  # Note: AWS feature is not yet available on registry.suse.com, hence the referral to registry.suse.de
+  image: "registry.suse.de/devel/unifiedcore/main/totest/containers/beta/uc/rke2/rke2-manifest:1.35.6"
```
ipetrov117, line 8: I am not a fan of this, but at least it enables the example's functionality. We could also just refer `registry.suse.com` and say that the example will be broken until the AWS feature is released. Let me know what you think.

davidcassany (approved):

dharmit (approved):

dharmit:
Separately, can we perhaps add some example that would show up on https://opensource.suse.com/elemental?

ipetrov117:
> Separately, can we perhaps add some example that would show up on https://opensource.suse.com/elemental?

I am working on an actual documentation for the feature as we speak.

atanasdinov (approved):

PR #590, opened 2026-09-09
Netboot Implementation

dbw7:
Introduces `--type netboot`

What this new type does is it generates the typical ISO along with a netboot directory. The netboot directory extracts all of the necessary artifacts needed for netboot to succeed. It is expected that the user will set up a web server to serve the netboot directory so it can be utilized at boot time. The generated contents look like so:

```
├── customized.iso
└── customized-netboot/   
    ├── boot/
    │   ├── grub2/grub.cfg  
    │   ├── initrd               
    │   └── vmlinuz 
    └── EFI/BOOT/
        ├── bootx64.efi          
        ├── grub.cfg             
        ├── grub.efi             
        └── MokManager.efi 
```

Inside of the `install.yaml` the user will have to specify the URL where the ISO will be accessible at. We expect that the user will take the elemental customized ISO and then serve it in some way but provide the expected URL before hand:
```
iso:
  device: "/dev/sda"
  netbootURL: http://localhost:8080/elemental.iso
```
This URL can be `http://`, `https://`, and `nfs://`

For NFS the user has to set up their own NFS handling, but the RAM usage is significantly lower during boot time.

This update requires a change in the OBS image to include specific artifacts to support this properly.


pkg/installer/media.go
```diff
@@ -69,6 +71,21 @@ func (m MediaType) String() string {
...
+}
+
+func (m MediaType) Ext() string {
+	switch m {
+	case ISO:
+		return "iso"
+	case Disk:
+		return "raw"
+	case Netboot:
+		return "iso"
```
dharmit: You could combine two like:
```go
case ISO, Netboot:
    return "iso"
case Disk:
    return "raw
```

internal/cli/action/customize.go
```diff
@@ -95,7 +97,11 @@ func Customize(ctx context.Context, cmd *cli.Command) error {
 
 func resolveOutputPaths(fs vfs.FS, args *cmdpkg.CustomizeFlags) (imagePath, configPath string) {
 	imagePath = args.OutputPath
-	imageName := fmt.Sprintf("image-%s.%s", time.Now().UTC().Format("2006-01-02T15-04-05"), args.MediaType)
+	ext := args.MediaType
+	if mType, err := installer.StringToMediaType(args.MediaType); err == nil {
```
dharmit: Please add a check for `err != nil` so that we could fail early when required. Without that, we fail quite late in the flow.

internal/cli/action/customize.go
```diff
@@ -147,13 +153,18 @@ func setupConfigManager(s *sys.System, configDir string, local bool) *config.Man
 func setupFileExtractor(ctx context.Context, s *sys.System, outDir config.Output, local bool) (extr *extractor.OCIFileExtractor, err error) {
 	const isoSearchGlob = "/iso/*default-iso*.iso"
 
-	if err := vfs.MkdirAll(s.FS(), outDir.ISOStoreDir(), vfs.DirPerm); err != nil {
+	storeDir := outDir.ISOStoreDir()
+	if cacheDir, cErr := os.UserCacheDir(); cErr == nil {
```
dharmit: Probably a check for `cErr != nil` before this would be helpful? Correct me if I'm wrong.

dharmit (changes requested):

internal/cli/action/customize.go
dbw7: This was a temporary convenience change (before the airgapping landed) but looking at it, I'm just gonna omit it for now.

davidcassany (approved):

internal/image/install/install.go
```diff
@@ -72,5 +72,6 @@ type RAW struct {
 }
 
 type ISO struct {
-	Device string `yaml:"device"`
+	Device     string `yaml:"device"`
+	NetbootURL string `yaml:"netbootURL" validate:"omitempty,http_url"`
```
ipetrov117, line 76: Wouldn't this type of validation block `nfs://` urls? IIUC, it allows only valid HTTP(s) urls.

pkg/deployment/deployment.go
```diff
@@ -226,6 +226,7 @@ type Partitions []*Partition
 type Disk struct {
 	Device     string     `yaml:"target,omitempty" validate:"disk_device_required,disk_device_exists"`
 	Partitions Partitions `yaml:"partitions" validate:"required,min=1,dive"`
+	NetbootURL string     `yaml:"netbootURL,omitempty"`
```
ipetrov117, line 229: I am having trouble locating where this is used. Is this a leftover, or serving as a placeholder for logic that is yet to come?

internal/cli/action/customize.go
```diff
@@ -93,10 +98,14 @@ func Customize(ctx context.Context, cmd *cli.Command) error {
 	return nil
 }
 
-func resolveOutputPaths(fs vfs.FS, args *cmdpkg.CustomizeFlags) (imagePath, configPath string) {
-	imagePath = args.OutputPath
-	imageName := fmt.Sprintf("image-%s.%s", time.Now().UTC().Format("2006-01-02T15-04-05"), args.MediaType)
+func resolveOutputPaths(fs vfs.FS, args *cmdpkg.CustomizeFlags) (imagePath, configPath string, err error) {
+	mType, err := installer.StringToMediaType(args.MediaType)
+	if err != nil {
+		return "", "", fmt.Errorf("parsing media type %s", args.MediaType)
```
ipetrov117, line 104: nit: Should we also wrap the downstream error message in here as well? That way we would have the full error chain.

pkg/bootloader/grub.go
```diff
@@ -535,6 +598,57 @@ func (g *Grub) installKernelInitrd(rootPath, espDir, subfolder string, extension
...
+	g.s.Logger().Info("Installing kernel/initrd")
+	entry := grubBootEntry{}
+
+	targetDir := filepath.Join(target, liveBootPath)
+	err := vfs.MkdirAll(g.s.FS(), targetDir, vfs.DirPerm)
+	if err != nil {
+		return entry, fmt.Errorf("creating netboot target dir: %w", err)
+	}
+
+	kernelPath, err := vfs.FindFile(g.s.FS(), rootPath, filepath.Join(liveBootPath, "*/*/vmlinuz*"))
```
ipetrov117, line 617: A probably ignorant/dumb question, but is this intended to function outside of `x86_64`? IIUC that is why `vfs.FindKernel` has the [kernel-patterns](https://github.com/dbw7/elemental/blob/netboot-imp/pkg/sys/vfs/vfs.go#L671-L677) slice, so that it can correctly locate the kernel paths for the different archs. 

I was wondering how does this transfer over to the netboot logic? Are we targeting just `x86_64` at the moment, or am I misreading the logic?

ipetrov117 (commented):
Great job on this @dbw7! 

Apart from the linter error that needs to be handled, everything looks good to me. I left out some minor comments mostly for my understanding.

pkg/bootloader/grub.go
dbw7: Yes this is intended to work on x86_64, I just forgot to test it completely but I've done some abstractions to vfs.FindKernel so now it's reusable for both netboot as well as its original usecase. Thank you for catching!

pkg/deployment/deployment.go
dbw7: Yes that was a leftover, thank you

internal/image/install/install.go
dbw7: You're correct, I had some mismatched states in my codebase, I added the correct validation for this

ipetrov117 (approved):

dharmit (approved):

atanasdinov (commented):
Looks great. One thing that I've been wondering is how to improve the UX.

Do we need to have `--type netboot`? Can we assume that by providing `netbootURL` (always paired with `--type iso`) it would be enough and clear, or is that a bad idea?

atanasdinov (approved):

